PR #6, opened 2025-06-18
Docs: Correct 'provisioner' to 'provider' in networking section

0x00-A:
The documentation incorrectly used the term 'provisioner' when referring to the entity that manages the VM and its networking (e.g., VirtualBox).

This change replaces the incorrect term with 'provider' to accurately reflect Vagrant's terminology and avoid confusion for new users. A provider manages the machine's lifecycle, whereas a provisioner configures software within the machine.


mariadb-stefan-hinz:
Thanks for the fix!

PR #8, opened 2025-06-19
unquote URL links

grooverdan:
They aren't rendered once quoted.

performed with:

git grep -lE '`\[[a-z-]+\]\('| xargs sed -i -E 's/`(\[[^]]+\]\([^)]+\))`/\1/g'


mariadb-stefan-hinz:
> @mariadb-stefan-hinz maybe late - but was quoting within `[]`correctly rendered like [`this`](link)? It can be changed to this.

I think we're done fixing those links, thanks to your contribution. Or are you suggesting other changes?

PR #25, opened 2025-07-03
node.js connector 3.4.4 release update

rusher:
This is the documentation release part of [TODO-5437](https://jira.mariadb.org/browse/TODO-5437)


rusher:
for @dbart

PR #27, opened 2025-07-05
Fix duplicate words

philip:
Fix duplicate words in documentation (e.g., "the the" -> "the")


philip:
Thanks for merging @mariadb-stefan-hinz! I _think_ I only touched doubled words despite having impulses to update more along the way. Given the number of files changed I suspect the merge conflicts came after the PR since this repo is so heavily edited, or maybe not :)

PR #39, opened 2025-07-16
DOCS-4166 Fix an anchor link reference

mariadb-YuchenPei:
Should be lower case #operations


mariadb-tauseefkhan:
@mariadb-YuchenPei The markdown link you modified is returning 404.

mariadb-stefan-hinz (approved):

PR #114, opened 2025-09-16
Docs/hashicorp key management.md

mariadb-stefangenov:


tools/mariadb-enterprise-operator/plugins/hashicorp-key-management.md
```diff
@@ -0,0 +1,229 @@
...
+        bind-address=*
+        default_storage_engine=InnoDB
+        binlog_format=row
+        innodb_autoinc_lock_mode=2
+        innodb_buffer_pool_size=800M
+        max_allowed_packet=256M
+
+        plugin_load_add = hashicorp_key_management
+        hashicorp-key-management-vault-url=http://vault-0.vault-internal.vault.svc.cluster.local:8200/v1/mariadb # Change to your Vault URL.
+        hashicorp-key-management-token=EXAMPLE_TOKEN # This needs to be changed to your token
```
mariadb-mmontes, line 82: ⚠️ We cannot  seed a Vault token in a ConfigMap, specially if it has elevated permissions

Alternatively, we can supply the HashiCorp configuration in a `Secret`,  and use the `volumes` and `volumeMounts` to mount it under ` /etc/mysql/mariadb.conf.d/`

tools/mariadb-enterprise-operator/plugins/hashicorp-key-management.md
```diff
@@ -0,0 +1,229 @@
...
+        [mariadb]
+        bind-address=*
+        default_storage_engine=InnoDB
+        binlog_format=row
+        innodb_autoinc_lock_mode=2
+        innodb_buffer_pool_size=800M
+        max_allowed_packet=256M
+
+        plugin_load_add = hashicorp_key_management
+        hashicorp-key-management-vault-url=http://vault-0.vault-internal.vault.svc.cluster.local:8200/v1/mariadb # Change to your Vault URL.
```
mariadb-mmontes, line 81: We should include a `hashicorp-key-management-vault-ca` to establish the TLS trust:
https://mariadb.com/docs/server/security/securing-mariadb/encryption/data-at-rest-encryption/key-management-and-encryption-plugins/hashicorp-key-management-plugin#hashicorp-key-management-vault-ca

tools/mariadb-enterprise-operator/plugins/hashicorp-key-management.md
```diff
@@ -0,0 +1,229 @@
...
+    Note: Here you should use the `path` we chose in the previous step.
+    ```sh
+    vault kv put /mariadb/1 data="$(openssl rand -hex 32)"
+    vault kv put /mariadb/2 data="$(openssl rand -hex 32)"
+    ```
+
+3. **(Optional) Create An Authentication Token.**
+    This step can be skipped if you want to use your own token. Consult with a Vault administrator regarding this. If you want to create a new one:
+    ```sh
+    vault token create
```
mariadb-mmontes, line 42: Although the [plugin documentation](https://mariadb.com/docs/server/security/securing-mariadb/encryption/data-at-rest-encryption/key-management-and-encryption-plugins/hashicorp-key-management-plugin) says:
>  Authentication is done using the Hashicorp Vault's token authentication method

This will be a static token, something not desirable by any security team. We could stick to this docs, but refer to the other available Vault authentication methods, which allow you to generate a short-lived Vault token. For example, the approle one can generate a Vault token based on a acces id and secret key:
https://developer.hashicorp.com/vault/docs/auth/approle

So the user could authenticate with any of the [supported methods by Vault](https://developer.hashicorp.com/vault/docs/auth), and obtain a Vault token, rather than creating a static token.

tools/mariadb-enterprise-operator/plugins/hashicorp-key-management.md
```diff
@@ -0,0 +1,229 @@
...
+    ```
+    You will see output similar to:
+    ```
+    Key                  Value
+    ---                  -----
+    token                EXAMPLE_TOKEN
+    token_accessor       9sHvwKiddpCT2PaplcHHRAnv
+    token_duration       ∞
+    token_renewable      false
+    token_policies       ["root"]
```
mariadb-mmontes, line 52: Ideally, we should have a custom policy that allows all operations under the `/mariadb/*` keys, rather than in the entire Vault.

tools/mariadb-enterprise-operator/plugins/hashicorp-key-management.md
```diff
@@ -0,0 +1,229 @@
...
+    ```
+    This will start re-encrypting data.
+
+2. **Monitor Re-Encryption.**
+    ```sh
+    kubectl run mariadb-connect --rm -it --image=mariadb:11.4 -- bash -c "mariadb -u root -p'MariaDB11!' --ssl=false -h mariadb"
+    ```
+    If you check the encrpytion status again:
+    ```sql
+    SELECT * from information_schema.INNODB_TABLESPACES_ENCRYPTION;
```
mariadb-mmontes, line 198: Could we share the output of this query?

Additionally, as described in [this blog](https://mariadb.com/resources/blog/how-to-implement-encryption-at-rest-using-hashicorp-vault-and-mariadb/), we can show that the `/var/lib/mysql/testdb/*.ibd` is actually encrypted.

tools/mariadb-enterprise-operator/plugins/hashicorp-key-management.md
```diff
@@ -0,0 +1,229 @@
...
+    ```
+    You should see `CURRENT_KEY_VERSION` column start getting updated to point to the new key version.
```
mariadb-mmontes, line 200: Show an example output of this `CURRENT_KEY_VERSION`

tools/mariadb-enterprise-operator/plugins/hashicorp-key-management.md
mariadb-mmontes: I know this is done like this [in the blog](https://mariadb.com/resources/blog/how-to-implement-encryption-at-rest-using-hashicorp-vault-and-mariadb/), but it is far from ideal.

Besides, the plugin is not configuring any authentication method nor anything global in vault, it just reads keys under a kv path, so their permissions should be scoped to `/mariadb/*`.

tools/mariadb-enterprise-operator/plugins/hashicorp-key-management.md
```diff
@@ -0,0 +1,229 @@
...
+    ```
+
+    At this point, you can check the encryption status:
+    ```sql
+    SELECT * from information_schema.INNODB_TABLESPACES_ENCRYPTION;
+    ```
+
+    If you create a new database and then table, the above query should return additional information about them.
+
+## Day-2 Ops
```
mariadb-mmontes, line 180:
```suggestion
## Day-2 Operations
```

tools/mariadb-enterprise-operator/plugins/hashicorp-key-management.md
```diff
@@ -0,0 +1,229 @@
...
+
+### Rotating Secrets.
```
mariadb-mmontes, line 182: This is way better than what the file encryption plugin:
https://mariadb.com/resources/blog/mariadb-encryption-tde-using-mariadbs-file-key-management-encryption-plugin/

Do you see any benefit of supporting the file encryption plugin at this point?

mariadb-mmontes (changes requested):
This is looking good! Left some comments.

tools/mariadb-enterprise-operator/plugins/pam.md
```diff
@@ -179,11 +179,11 @@ metadata:
 spec:
   rootPasswordSecretKeyRef:
     name: mariadb
-    key: password
+    key: root-password
```
mariadb-mmontes, line 182: I guess we want to stick to `examples/manifests/config` ?

mariadb-mmontes (changes requested):

tools/mariadb-enterprise-operator/plugins/hashicorp-key-management.md
mariadb-mmontes: We could explore extending the operator to support `myCnfSecretKeyRef`  in further iterations, to achieve this in a more native way.

tools/mariadb-enterprise-operator/plugins/hashicorp-key-management.md
mariadb-stefangenov: Yeah, the dev server didn't care for policies :smile: But now since we want a TLS example I had to spin up a standalone instance and a custom policy was a must.

tools/mariadb-enterprise-operator/plugins/hashicorp-key-management.md
mariadb-stefangenov: I feel like we are lacking in options. The file key management plugin is not production ready, while vault is great, but it's the **only** option right now.

At the same time, for some customers that file key management plugin could be a good middle ground.

tools/mariadb-enterprise-operator/plugins/hashicorp-key-management.md
mariadb-stefangenov: Ah, tunnel visioned here. I was literally working with Secret Mounts for ldap, yet forgot I can use them for this too. Thanks, that is perfect

tools/mariadb-enterprise-operator/plugins/hashicorp-key-management.md
Michaelpalacce: approle will be a problem too, unless it has 0 ttl, which wouldn't really be any different from tokens I believe.

tools/mariadb-enterprise-operator/plugins/hashicorp-key-management.md
```diff
@@ -0,0 +1,303 @@
...
+    Your new token is: `EXAMPLE_TOKEN`.
+
+4. **Create A `Secret` For the vault token.**
+    Now that you've either created a new token, or are using an existing one, we need to create a secret with it.
+    ```sh
+    export TOKEN="EXAMPLE_TOKEN"
+    kubeclt create secret generic mariadb-vault-token --from-literal=token="$TOKEN"
+    ```
+
+5. **Create a Secret for the ca.**
```
mariadb-mmontes, line 78:
```suggestion
5. **Create a Secret for the Certificate Authority (CA) used to issue the Vault certificate.**
```

tools/mariadb-enterprise-operator/plugins/hashicorp-key-management.md
```diff
@@ -0,0 +1,303 @@
...
+    For which certificate is needed, consult [the docs](../../../server/security/securing-mariadb/encryption/data-at-rest-encryption/key-management-and-encryption-plugins/hashicorp-key-management-plugin.md#hashicorp-key-management-vault-ca)
```
mariadb-mmontes, line 79:
```suggestion
    For further information, consult [the docs](../../../server/security/securing-mariadb/encryption/data-at-rest-encryption/key-management-and-encryption-plugins/hashicorp-key-management-plugin.md#hashicorp-key-management-vault-ca)
```

tools/mariadb-enterprise-operator/plugins/hashicorp-key-management.md
```diff
@@ -0,0 +1,303 @@
...
+    | innodb_undo002  |                 1 |               1 |                   1 |              1 |                    0 |
+    | innodb_undo003  |                 1 |               1 |                   1 |              1 |                    0 |
+    | mysql/innodb_ta |                 1 |               1 |                   1 |              1 |                    0 |
+    | mysql/innodb_in |                 1 |               1 |                   1 |              1 |                    0 |
+    | mysql/gtid_slav |                 1 |               1 |                   1 |              1 |                    0 |
+    | mysql/transacti |                 1 |               1 |                   1 |              1 |                    0 |
+    | my_db/people    |                 1 |               1 |                   1 |              1 |                    0 |
+    +-----------------+-------------------+-----------------+---------------------+----------------+----------------------+
+    ```
+    Note: The above query is truncated, so it will be readable in the docs. In reality, you will see a few more columns.
```
mariadb-mmontes, line 220:
```suggestion
    Note: The above query is truncated. In reality, you will see a few more columns.
```

tools/mariadb-enterprise-operator/plugins/hashicorp-key-management.md
```diff
@@ -0,0 +1,303 @@
...
+Make sure when rotating the token, to do so in advance of the token expiring.
+
+1. **Acquire a new token and update the secret.**
+
+    ```sh
+    export TOKEN="EXAMPLE_TOKEN"
+    kubeclt create secret generic mariadb-vault-token --from-literal=token="$TOKEN"
+    ```
+
+2. **Restart The MariaDB StatefullSet.**
```
mariadb-mmontes, line 270:
```suggestion
2. **Restart The MariaDB StatefulSet.**
```

tools/mariadb-enterprise-operator/plugins/hashicorp-key-management.md
```diff
@@ -0,0 +1,303 @@
...
+    MariaDB will continue using the old token until the statefulset is restart. If the name of the database CRD is `mariadb` (as per the example above), then:
```
mariadb-mmontes, line 271:
```suggestion
    MariaDB will continue using the old token until the StatefulSet is restarted. If the name of the `MariaDB` CR  is `mariadb` (as per the example above), then:
```

tools/mariadb-enterprise-operator/plugins/hashicorp-key-management.md
```diff
@@ -0,0 +1,303 @@
...
+    ```
+
+## Known Issues/Limitations
+
+### **Vault Not Being Accessible Will Result In MariaDB Not Working**
+
+As MariaDB uses vault to fetch it's decryption key, in the case where vault is not accessible due to it being down, network issues, secret
+not existing, will result in MariaDB not being able to fetch the decryption key and hence stop working. While the Hashicorp plugin has a
+configurable cache, that should be set and will result in MariaDB still working for a few seconds to minutes, depending on configuration,
+the cache is not reliable as it's ephemeral and short lived.
```
mariadb-mmontes, line 283:
```suggestion
As MariaDB uses Vault to fetch it's decryption key, in case that Vault becomes unavailable, it will result in MariaDB not being able to fetch the decryption key and hence stop working. While the Hashicorp plugin has a configurable cache, that should be set and will result in MariaDB still working for a few seconds to minutes, depending on configuration, the cache is not reliable as it's ephemeral and short lived.
```

tools/mariadb-enterprise-operator/plugins/hashicorp-key-management.md
```diff
@@ -0,0 +1,303 @@
...
+
+    **mariadb-vault.yaml**
+    ```yaml
+    ---
+    apiVersion: v1
+    kind: Secret
+    metadata:
+      name: mariadb # Used to hold the mariadb and root user passwords
+      labels:
+        enterprise.mariadb.com/watch: ""
```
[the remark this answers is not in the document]
mariadb-stefangenov: Hmm, but in this case it should be ok, no? It's also part of the examples.

tools/mariadb-enterprise-operator/plugins/hashicorp-key-management.md
```diff
@@ -0,0 +1,303 @@
...
+        encrypt_tmp_disk_tables = ON
+        encrypt_tmp_files = ON
+        encrypt_binlog = ON
+        aria_encrypt_tables = ON
+
+        innodb_encryption_threads = 4
+        innodb_encryption_rotation_iops = 2000
+
+      env:
+        - name: VAULT_TOKEN # This is where our token is defined!
```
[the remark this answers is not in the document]
mariadb-stefangenov: I have no idea, I mentioned it in one of the commits I believe but it's basically reverse engineering (in this case from a log that the plugin adds).

tools/mariadb-enterprise-operator/plugins/hashicorp-key-management.md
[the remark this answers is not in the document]
mariadb-mmontes: True, for one moment I got confused with the Vault token rotation

tools/mariadb-enterprise-operator/plugins/hashicorp-key-management.md
mariadb-mmontes: We should raise this with Egor when delivering the docs.

tools/mariadb-enterprise-operator/plugins/hashicorp-key-management.md
mariadb-mmontes: It is clear the way it is

PR #162, opened 2025-10-30
Parameter name typo fix

hartmut-mariadb:
The parameter name for max retries is 

`hashicorp-key-management-max-retries`

not just 

`hashicorp-key-management-retries`

```
MariaDB [(none)]> show variables like "hashi%retries";
+--------------------------------------+-------+
| Variable_name                        | Value |
+--------------------------------------+-------+
| hashicorp_key_management_max_retries | 3     |
+--------------------------------------+-------+
```


iangilfillan:
Thanks, this is incorrect in a number of places, including the server itself: https://github.com/MariaDB/server/blob/main/plugin/hashicorp_key_management/hashicorp_key_management.cnf

PR #190, opened 2025-11-13
Add a codespell check in GH CI

fauust:


mariadb-stefan-hinz:
@fauust Thanks for a first stab at implementing this! Adding @mariadb-tauseefkhan to this thread, particularly since I had created an implementation ticket for him already. :)
From my layman's perspective, seems like running the checker on the diff only seems like the best option.

fauust:
Ok, I quickly changed a bit the PR using the official codespell action and replicating what you already do with link check (which is nice!). Let me know what you think.

PR #199, opened 2025-11-17
Fix typo in maximum length of text data type

idarlm:
Fixes a small typo in the docs for the TEXT data type

### DIFF
--- a/server/reference/data-types/string-data-types/blob.md
+++ b/server/reference/data-types/string-data-types/blob.md
@@ -8,7 +8,7 @@ BLOB[(M)]
 
 ## Description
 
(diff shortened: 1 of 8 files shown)


idarlm:
I see that mediumtext and longtext also have the same issue, i can update the PR to fix those as well this afternoon

iangilfillan:
Thanks, yes, looks like a failure when the import was done from the KB, so these probably exist in multiple places. Please go ahead and update this PR to include the others you're aware of.

idarlm:
I have fixed the occurrences  I found across the string docs. I saw that `tinytext.md` used UTF-8 superscript so I used that across all the other articles as well, hope that's fine.

PR #294, opened 2026-01-13
Change wsrep_sst_method to wsrep_osu_method

DNS320:
Hello

First of all, thank you for the excellent MariaDB documentation! It is a pleasure to read.

I think the mention of `wsrep_sst_method` in [Configuring Wsrep GTID Mode](https://mariadb.com/docs/galera-cluster/high-availability/using-mariadb-replication-with-mariadb-galera-cluster/configuring-mariadb-replication-between-two-mariadb-galera-clusters#configuring-wsrep-gtid-mode) should be changed to [`wsrep_osu_method`](https://mariadb.com/docs/galera-cluster/reference/galera-cluster-system-variables#wsrep_osu_method). This PR does just that.

Hope I could help & have a nice day!


sjaakola (approved):
a good one

PR #300, opened 2026-01-16
Update innodb-system-variables.md

DNS320:
Hello

I tested the `innodb_stats_on_metadata` setting and believe that the default value is not correctly described in the description. I also changed the possible values from 1/0 to ON/OFF.


mariadb-stefan-hinz (commented):
ON and 1 have the same effect. I'm fine with saying ON – I think it's easier to understand. However, we shouldn't remove what the default is (1, or ON). I'm adding that back.

PR #310, opened 2026-01-19
Update innodb-system-variables.md

DNS320:
Hello
This is an addition to https://github.com/mariadb-corporation/mariadb-docs/pull/300, as I forgot to mention why I deleted the word `default` from the description. The reason for this was that the correct default setting is `OFF` and not `ON`.
Since the reviewer wants to leave the standard text in the description, I simply moved it now to the `OFF` section.

Thank you for your work!


mariadb-stefan-hinz:
Thanks for fixing the default value! You're right, it's OFF, not ON.

PR #359, opened 2026-02-11
Fix gen settings reference

jhnwkmn:
These commits:
- Fixes a few issues.
- Adapts the script that generates a setting reference to the new directory and setting layout.
- Updates the settings reference.
-


mariadb-stefan-hinz:
Looks good. Thanks for improving and automating the docs!

PR #363, opened 2026-02-13
12.3.1 release notes missing an item

mariadb-YuchenPei:


mariadb-stefan-hinz:
Thanks for adding the missing feature!

PR #389, opened 2026-03-05
Update securing-connections-for-client-and-server.md

vuvova:


mariadb-stefan-hinz:
Thanks for improving the docs!

PR #444, opened 2026-04-08
docs: Updated some examples to show that they are potentially incomplete

mariadb-stefangenov:


mariadb-mmontes (commented):
Not entirely sure about this. I remember adding examples in PITR without `...`. Maybe assuming that they are Kustomize overlays ? 😄 

Whatever we do, we should be consistent across all docs

mariadb-stefangenov:
@mariadb-mmontes can you take anther look?

mariadb-mmontes (approved):
@mariadb-stefangenov LGTM! Really like the new SBOM section

PR #449, opened 2026-04-14
Update mariadb-audit-plugin-log-format.md

vuvova:
remove any references of "milliseconds", they are not supported, never were


dbart:
Merged. Thanks!

PR #451, opened 2026-04-15
Clarify `transaction_replay_safe_commit` behavior

markus456:
Clarified the behavior of `transaction_replay_safe_commit` and added examples of transactions that would result in side-effects if a replay did occur with `transaction_replay_safe_commit=false`.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -27,7 +27,7 @@ Instead of using a full internal URL, use the following syntax: `[Link Text]({al
 
 Example:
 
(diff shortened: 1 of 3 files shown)


markus456:
@mariadb-tauseefkhan @mariadb-stefan-hinz I don't know how that automation works but it seems to have added unrelated changes. Can you evaluate whether that behavior is correct? The link expansion seems strange.

mariadb-tauseefkhan:
@markus456 This has to do with the GitBook action https://github.com/mariadb-corporation/mariadb-docs/actions/workflows/expand-gitbook-aliases.yml. Let me fix it.

PR #452, opened 2026-04-15
Update use-database.md

zpeschke:
Minor grammar fix I came across while reading this web page.


mariadb-stefan-hinz:
Thanks for the fix!

PR #476, opened 2026-05-07
Clarify DDLs using statement format when binlog_format=row

bnestere:
There was some confusion about falling back to `STATEMENT` format when `binlog_format=row` for DDLs, so make it explicit that statement format is used (rather than leaving it at just "logged").


mariadb-stefan-hinz (approved):
Thanks for clarifying, Brandon!

PR #477, opened 2026-05-08
Minor grammatical change to text about PITR MariaDB backup

sjmudd:
Tiny suggested change.


mariadb-stefan-hinz (approved):
Thanks for fixing, Simon!

PR #479, opened 2026-05-08
MXS-6234 Add Keepalived tutorial

ekorh475:
Based off on the old Keepalive+MaxScale blog post. Various parts were updated, and the setup tested with current Keepalived version.


mariadb-stefan-hinz (approved):

mariadb-stefan-hinz (approved):
I'd suggest to say "MaxScale instances" rather than "MaxScales", but that's a minor. Otherwise, the tutorial looks good. Leaving it to Esa to change the minor issue I raised.

PR #490, opened 2026-05-14
help-tables: fix bootstrap-compat check and unbreak workflow

mariadb-ElijahOduba:
## Pull Request

Unbreaks the `Generate Help Tables SQL` workflow (broken on `main` since #484 merged) and makes the generator output `mariadbd --bootstrap`-compatible so @grooverdan can unblock [MDEV-39584](https://jira.mariadb.org/browse/MDEV-39584) / MariaDB/server#5068.

## Type of Change

- Bug fix

## Description

Three layered fixes:

1. **Workflow YAML**: the `Check server compatibility` step Daniel added in #484 has backslash-line-continued `bash -c '...'` inside `run: |` that GitHub's parser rejects with `line 50 syntax error` (passes `actionlint` and PyYAML, fails the real thing). Rewritten as a clean `sh -c '...'` block. This is why every push to `main` since 2026-05-13 has been a 0-second CI failure.
2. **`truncate_to_bytes`** sized for `help_topic.description` (TEXT, 65 KB) but `sql_bootstrap.h` caps each query at `MAX_BOOTSTRAP_QUERY_SIZE` = 20,000 bytes. Dropped the cap to 15,000 raw bytes so the escaped INSERT plus boilerplate fits.
3. **`escape_sql`** keeps `\n -> \\n`; comment updated to explain why (bootstrap parser is line-naive; it skips blank/`--`/`#` lines and treats any line ending in `;` as a statement terminator, without tracking string-literal state, so real newlines inside string literals would corrupt INSERTs).

`validate_sql.py` rewritten to parse statements (split on `;` outside strings) instead of lines, and to enforce the 20K-per-INSERT limit explicitly so this regression can't recur silently.

Verified end-to-end against the same Docker bootstrap step CI runs, plus a populated server returning correct `HELP 'SELECT' / 'COUNT' / 'GROUP BY'`. Fork CI green: https://github.com/mariadb-ElijahOduba/mariadb-docs/actions/runs/25868580315

### DIFF
--- a/.github/workflows/generate-help-tables.yml
+++ b/.github/workflows/generate-help-tables.yml
@@ -44,10 +44,11 @@ jobs:
 
       - name: Check server compatibility
         run: |
-          docker run --rm -v $PWD/help-tables:/docker-entrypoint-initdb.d mariadb:11.4 \
-            bash -c '(echo -e "create database mysql;\nuse mysql;"; \
-                      cat /usr/share/mariadb/mariadb_system_tables.sql /docker-entrypoint-initdb.d/fill_help_tables.sql) \
-                      | /usr/sbin/mariadbd --bootstrap'
+          docker run --rm -v "$PWD/help-tables:/help-tables" mariadb:11.4 sh -c '
+            { printf "create database mysql;\nuse mysql;\n";
+              cat /usr/share/mariadb/mariadb_system_tables.sql /help-tables/fill_help_tables.sql;
+            } | mariadbd --bootstrap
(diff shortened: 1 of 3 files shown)


grooverdan:
:heart: , thank you.
workflow - https://github.com/mariadb-ElijahOduba/mariadb-docs/actions/runs/25869262171
I've tested the generated artifact in the the server and it functions on bootstrap.

grooverdan (commented):
d

PR #509, opened 2026-05-22
MDEV-39610 - Modify user privileges in create-user SQL example

julienfritsch44:
Updated privileges granted to user 'joffrey' in SQL example to reflect INSERT instead of SELECT for certain connections.


julienfritsch44 (commented):
@stefanhinz ready to be reviewed

PR #512, opened 2026-05-26
⚠️ WIP ⚠️ Docs/enterprise operator release 26.06.0

mariadb-stefangenov:


tools/mariadb-enterprise-operator/security/fips.md
```diff
@@ -0,0 +1,110 @@
...
+- Amazon S3 compatible storage
+- Azure Blob Storage
+- The MariaDB Agent Sidecars
+- Health check probes
+
+By enforcing these specific curves, the operator guarantees that all its external communication over TLS is FIPS-compliant.
+
+### OpenSSL Configuration
+
+While the Go runtime's FIPS mode (`GODEBUG=fips140=on`) ensures that the Operator's standard library cryptographic routines are compliant, it does not affect other cryptographic libraries running inside the containers, most notably OpenSSL. The MariaDB server, MaxScale, and various system utilities rely on OpenSSL for their cryptographic operations.
```
mariadb-mmontes:
```suggestion
While the Go runtime's FIPS mode (`GODEBUG=fips140=on`) ensures that the Operator's standard library cryptographic routines are compliant, it does not affect other cryptographic libraries running inside the operand containers, most notably OpenSSL. The MariaDB server, MaxScale, and various system utilities rely on OpenSSL for their cryptographic operations.
```

mariadb-mmontes: > While the Go runtime's FIPS mode (`GODEBUG=fips140=on`) ensures that the Operator's standard library cryptographic routines are compliant , it does not affect other cryptographic libraries running inside the operand container

Feels like we are mixing the Go and OpenSSL topics here. Can we create a brief section before this one covering the Go crypto details, scoping it to the operator and prometheus exporters? 

This way, we can also dedicate this section for OpenSSL and operand containers in a more explicit way.

tools/mariadb-enterprise-operator/security/fips.md
```diff
@@ -0,0 +1,110 @@
...
+
+To ensure end-to-end FIPS compliance within the Pod, OpenSSL must also be configured to operate in FIPS mode. **The Operator relies on the underlying host operating system to be FIPS compliant and to provide the necessary validated OpenSSL FIPS provider modules.**
```
mariadb-mmontes: > The Operator relies on the underlying host operating system

The operator itself does not rely on the host operating system, this only applies to operand containers. See my previous comment regarding having separate Go + OpenSSL sections for better clarity

tools/mariadb-enterprise-operator/security/fips.md
```diff
@@ -0,0 +1,110 @@
...
+  config:
+    env:
+      - name: GODEBUG
+        value: fips140=on
+  # [...]
+```
+
+## Implications of FIPS Mode
+
+When FIPS mode is enabled, the Go runtime enforces strict rules on the cryptographic algorithms that can be used. The MariaDB Operator ensures that all its communications are FIPS-compliant.
```
mariadb-mmontes:
```suggestion
When FIPS mode is enabled, the Go runtime enforces strict rules on the cryptographic algorithms that can be used. The MariaDB Enterprise Operator ensures that all its communications are FIPS-compliant.
```

tools/mariadb-enterprise-operator/security/fips.md
```diff
@@ -0,0 +1,110 @@
...
+- `P-521`
+- `P-384`
+- `P-256`
+
+This enforcement applies to communication with:
+- Kubernetes API Server
+- Amazon S3 compatible storage
+- Azure Blob Storage
+- The MariaDB Agent Sidecars
+- Health check probes
```
mariadb-mmontes: Probes don't have TLS, as Kubernetes doesn't have support to configure a custom TLS trust chain for them. Probe server runs in a different port in the agent, without TLS (5566), whereas the agent API does have TLS (5555)

tools/mariadb-enterprise-operator/security/fips.md
```diff
@@ -0,0 +1,110 @@
...
+The MariaDB Enterprise Operator for Kubernetes can be configured to operate in a FIPS 140-3 compliant mode. This ensures that all cryptographic operations performed by the operator adhere to the strict standards required by FIPS.
+
+## Activating FIPS Mode
+
+To activate FIPS mode, you must set the `GODEBUG` environment variable for the operator controller. When installing via Helm, you can achieve this by setting the `extraEnv` value.
+
+Create a `values.yaml` file with the following content:
+
+```yaml
+extraEnv:
```
mariadb-mmontes, line 16: For phase 1 I think this would be enough, simple.

For phase 2, I am anticipating the following values:

```yaml
fips:
  enabled: true # adds GODEBUG=fips140=on
  tlsMode: true # adds an additional env or a flag to enforce TLS client auth
```

The only downside I see about not introducing the `fips.enabled=true` now is that a user that is migrating might end up with a duplicate `GODEBUG`, one leftover one from `extraEnv` and the one managed by us internally. This doesn't break the deployment I believe, so I believe it should be fine. Unless I am missing something, I think we can leave it simple as it is.

tools/mariadb-enterprise-operator/security/fips.md
mariadb-mmontes: >  cryptographic routines are compliant

I would reinforce, at least once, that the reason why they are compliant is that they use NIST-approved cryptographic modules modules linking to the NIST CMVP section, not just adhering to the recommended algorithms.

tools/mariadb-enterprise-operator/security/fips.md
```diff
@@ -0,0 +1,110 @@
...
+```sh
+helm install mariadb-enterprise-operator mariadb-enterprise-operator/mariadb-enterprise-operator \
+  -f values.yaml
+```
+
+This will inject the required environment variable into the operator's Pod to enable FIPS-compliant mode.
+
+### OpenShift
+
+If you have deployed the operator on OpenShift via OperatorHub, you need to edit the `Subscription` for the MariaDB Enterprise Operator to inject the environment variable. After deploying the operator, edit the subscription and add the `config` section to the spec:
```
mariadb-mmontes:
```suggestion
If you have deployed the operator on OpenShift, you need to edit the `Subscription` for the MariaDB Enterprise Operator to inject the environment variable. After deploying the operator, edit the subscription and add the `config` section to the spec:
```
OperatorHub does not include certified operators.

mariadb-mmontes (changes requested):
Reviewed FIPS-related documentation. Overall looks good, left a couple of suggestions.

tools/mariadb-enterprise-operator/topologies/replication.md
```diff
@@ -258,6 +258,35 @@ The steps involved in the switchover operation are:
...
+    size: 1Gi
+  replicas: 3
+  replication:
+    enabled: true
+    primary:
+      autoSwitchoverOnGracefulShutdown: false
+```
+
+{% hint style="warning" %}
+The switchover on graceful shutdown is executed in a `preStop` hook within the primary `Pod`. Consequently, there are no logs available for this operation. However, the operator emits Kubernetes events that can be inspected to observe the progress of the switchover. You can query these events using `kubectl get events` or watch them in real-time with `kubectl get events -w`.
```
mariadb-mmontes: Please add an output example reflecting the events during a complete switchover with maxscale.

mariadb-mmontes (changes requested):
FIPS phase 1 looks good. Left a suggestion for graceful shutdown events. 

We still need to cover multi-cluster topology.

tools/mariadb-enterprise-operator/migrations/migrate-graceful-shutdown-switchover-26-06.md
```diff
@@ -0,0 +1,38 @@
+# Migrate to Graceful Shutdown Switchover in 26.06
```
mariadb-mmontes: Great to have a dedicated guide for this. I'll be linking it in the general 26.06 migration guide.

tools/mariadb-enterprise-operator/topologies/assets/multi-cluster.drawio
```diff
@@ -0,0 +1,225 @@
+<mxfile host="app.diagrams.net" agent="Mozilla/5.0 (X11; Linux x86_64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/143.0.0.0 Safari/537.36">
```
mariadb-mmontes, line 1: This file should be renamed.

tools/mariadb-enterprise-operator/topologies/multi-cluster.md
```diff
@@ -0,0 +1,1004 @@
+---
+description: >-
+  Explains how to deploy and operate multi-cluster MariaDB Enterprise
+  topologies, replicating data between clusters across multiple Kubernetes
+  clusters or within a single one for high availability and disaster recovery.
+---
+
+# Multi-cluster
```
mariadb-mmontes, line 8: We should be adding a banner here saying it is in tech preview.

tools/mariadb-enterprise-operator/topologies/multi-cluster.md
```diff
@@ -0,0 +1,1004 @@
...
+    "mariadb-eu-central-1": "Replica"
+  }
+}
+```
+
+The `PrimaryReplica` Pod (`mariadb-eu-central-0`) replicates from the primary cluster's primary Pod. The `gtidCurrentPos` shows both domain `0` (replicated from the primary cluster) and domain `1` (its own cluster's transactions).
+
+### Scenarios
+
+The following sections describe the different scenarios supported by the multi-cluster feature. All scenarios are available in the [examples catalog](../examples-catalog.md).
```
mariadb-mmontes, line 403: Thanks for the reminder, added TODO to update examples catalog.

mariadb-mmontes (changes requested):
Thanks for taking care of the 26.06. I've identified some minor improvements that I'll be addressing myself. We also need to add:
- Migration guide for 26.06, linking to the specific guide for switchover on graceful shutdown you've built.

tools/mariadb-enterprise-operator/topologies/multi-cluster.md
```diff
@@ -7,6 +7,9 @@ description: >-
 
 # Multi-cluster
 
+{% hint style="info" %}
+Multi-cluster topology is available as a **Tech Preview**. It is not recommended for production use yet.{% endhint %}
```
mariadb-stefangenov: Move `{% endhint %}` to the bottom, not sure if it's accepted like this?

tools/mariadb-enterprise-operator/updates/update-26.06.md
```diff
@@ -0,0 +1,80 @@
...
+apiVersion: enterprise.mariadb.com/v1alpha1
+kind: MariaDB
+metadata:
+  name: mariadb-repl
+spec:
+  updateStrategy:
++   autoUpdateDataPlane: true
+```
+
+- First of all, the CRDs must be updated to `26.6.0`:
```
mariadb-stefangenov, line 23: Put this above:

```
- The [data-plane](../topologies/data-plane.md) must be updated to the `26.6.0` version. You must set `updateStrategy.autoUpdateDataPlane=true` in your `MariaDB` resources before updating the operator. Then, once updated, the operator will also update the data-plane based on its version:
```

PR #526, opened 2026-05-29
Clarify Connector/Python 2.0 is an RC; 1.1 remains GA

rusher:
MariaDB Connector/Python 2.0 is currently a **Release Candidate (RC)**, while
**1.1 is the latest stable (GA) release**. Because 2.0 is a pre-release, a plain
`pip install mariadb` installs the latest GA (1.1) — to get 2.0 you must pass
`--pre`. The docs were written as though 2.0 were GA, so the install commands
(`pip install mariadb`, `pip install mariadb[binary,pool]`, etc.) would have
silently installed 1.1 instead of the intended 2.0.

This PR makes the version status explicit and corrects the install commands.

### DIFF
--- a/connectors/connectors-quickstart-guides/connector-python-guide.md
+++ b/connectors/connectors-quickstart-guides/connector-python-guide.md
@@ -29,22 +29,37 @@ Before installing MariaDB Connector/Python, ensure you have:
 
 #### 2. Installation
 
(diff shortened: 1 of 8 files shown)


mariadb-stefan-hinz:
Thanks for the clarifications re. C/Python 1.1 vs. 2.0!

PR #530, opened 2026-05-30
Replaced all mentions of xtrabackup_checkpoints with mariadb_backup_c…

Centaro0:
…heckpoints

Replaced all mentions of xtrabackup_checkpoints with mariadb_backup_checkpoints as the name of the file changed at some point.

### DIFF
--- a/server/server-usage/backup-and-restore/mariadb-backup/incremental-backup-and-restore-with-mariadb-backup.md
+++ b/server/server-usage/backup-and-restore/mariadb-backup/incremental-backup-and-restore-with-mariadb-backup.md
@@ -1,7 +1,7 @@
 ---
 description: >-
   Complete mariadb-backup incremental guide: --backup/--prepare syntax, LSN
-  xtrabackup_checkpoints, --incremental-basedir, and --copy-back/--move-back
+  mariadb_backup_checkpoints, --incremental-basedir, and --copy-back/--move-back
   restore.
 ---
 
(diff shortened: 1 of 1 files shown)


mariadb-stefan-hinz:
Verified against the mariadb-server source: the claim is correct.

The rename was made by commit `b314f7b6420` (2023-03-10) for [MDEV-18931 "Rename Mariabackup's xtrabackup_\* files to mariadb_backup_\*"](https://jira.mariadb.org/browse/MDEV-18931), which renamed `xtrabackup_checkpoints` → `mariadb_backup_checkpoints` (and several sibling files). The new name is what `mariadb-backup` writes today; the old name is kept only as a backward-compatibility read fallback in `extra/mariabackup/xtrabackup.cc` (and in the WSREP SST script for one-shot migration).

One nuance worth knowing: the rename only landed in **11.1+**. It is *not* in 10.5, 10.6, 10.11, or 11.0. Users on the still-maintained 10.6 / 10.11 LTS branches will still see `xtrabackup_checkpoints` in their backup directory.

Two ways to go from here:

1. **Approve as-is.** The new name is correct for current MariaDB and the page isn't version-tagged.
2. **Ask for a small footnote** for 10.6 / 10.11 readers, e.g.:
   > In MariaDB 11.0 and earlier this file was named `xtrabackup_checkpoints`. The current name is recognized as a fallback for backups taken with older versions.

Not a blocker either way.

mariadb-stefan-hinz:
Thanks for your contribution! After merging, I'll add a note to the page:
In MariaDB 11.0 and earlier this file is named `xtrabackup_checkpoints`. The new name is recognized as a fallback for backups taken with older versions.

PR #534, opened 2026-06-02
fix group_concat_max_len max

byteplow:
group_concat_max_len max was changed by the latest minor releases


mariadb-stefan-hinz:
image
Thanks for fixing, byteplow!

PR #544, opened 2026-06-04
DOCS-5614 Retag bnf code blocks in Syntax sections (prep)

mariadb-stefan-hinz:
## Summary

- Retags the first code block under every `## Syntax` heading from `` ```sql `` to `` ```bnf `` across **799 files**.
- Preparatory work for [DOCS-5614 (Railroad-diagram authoring)](https://mariadbcorp.atlassian.net/browse/DOCS-5614) — the Railroad-diagram tool chain needs to know which code blocks are grammars vs. example SQL, and the existing markup didn't distinguish them.

## Why

The MariaDB docs convention is that `## Syntax` contains an EBNF/BNF grammar, while runnable SQL examples go under `## Examples` or `## Description`. Today only 13 reference pages tag the grammar block as `` ```bnf ``; the other ~800 use `` ```sql ``, making the source markup ambiguous. This sweep brings the markup in line with the established convention so that downstream tools (the Railroad-diagram generator, the help-tables extractor, the DX-project Skill files, etc.) can rely on the tag.

## Conservative scope

Only the **first** `` ```sql `` block inside each `## Syntax` section is retagged. A small number of pages use nested `### subsections` inside `## Syntax` that mix grammars with inline error/example output — most notably:

- `server/reference/sql-statements/transactions/xa-transactions.md`
- `server/reference/sql-statements/programmatic-compound-statements/for.md`
- `server/reference/sql-statements/data-manipulation/changing-deleting-data/delete.md`
- `server/ha-and-performance/optimization-and-tuning/optimizer-hints/expanded-optimizer-hints.md`
- `server/reference/sql-functions/special-functions/json-functions/json_table.md`

For those, the conservative approach leaves the secondary blocks as `` ```sql `` and we'll do a follow-up manual review pass to identify which additional blocks are genuinely BNF (and which are examples that should be moved to a proper `## Examples` section as docs hygiene).

## How it was done

A small Python script (~80 lines, regex-free, runs as a state machine over each Markdown file) traversed every `.md` file in the repo. The script tracks whether the parser is currently inside a `## Syntax` section vs. some other `## ` section, whether it's inside a code block, and whether the first `` ```sql `` fence in the current Syntax section has already been retagged. The retag only happens for the first eligible fence per section. The script source can be added to the repo if it would be useful for future similar sweeps.

## Diff shape

- **799 files changed** — all under-tagged grammar blocks now correctly use `` ```bnf ``.
- **Per-file diff is trivial**: a single line changes from `` ```sql `` to `` ```bnf `` near the top of the page (one or two lines below `## Syntax`).
- **No prose, examples, or syntax content was modified** — only the language tag on one code fence per page.

## Test plan
- [ ] Spot-check rendered output in GitBook preview on a few representative pages (e.g. `select.md`, `create-index.md`, `xa-transactions.md`) to confirm the retag doesn't break rendering.
- [ ] Confirm `codespell` and `lychee` pass (this change is pure tag edits, so they should be no-ops).
- [ ] Sanity-check a multi-block case like `xa-transactions.md` to verify only the first fence was retagged.


mariadb-stefan-hinz:
Not all of the syntax blocks are real BNF (Backus-Naur Form) diagrams, but most of the ones that weren't syntax-tagged 'bnf' were tagged 'sql' instead, which isn't true either. Using 'bnf' for those makes it clear where on a page the syntax is stated. Also, 'bnf' syntax highlighting is fairly neutral, while 'sql' highlighting can make syntax blocks look weird when the syntax isn't SQL.

PR #558, opened 2026-06-10
Added a section to clarify plugin maturity guidelines.

gkodinov:
Added a section to clarify the maturity guidelines for plugins. 
It deals with the release cycles and suggests maturity levels for each state.

See the discussion in https://github.com/MariaDB/server/pull/5203.

### DIFF
--- a/server/reference/product-development/plugin-development/development-writing-plugins-for-mariadb.md
+++ b/server/reference/product-development/plugin-development/development-writing-plugins-for-mariadb.md
@@ -71,6 +71,42 @@ maria_declare_plugin(example)
 maria_declare_plugin_end;
 ```
 
+## Maturity Guidelines For Plugins
+
+### Plugin Maturity At a Galnce
+
+```mermaid
+stateDiagram-v2
+    state "Release Candidate" as RC
+    state "General Availability" as GA
+    state "Experimental" as Exp {
+      [*] --> Alpha
+      [*] --> Beta
+      Alpha --> Beta
+      Beta --> [*]
+      [*] --> [*]
+    }
+    Exp --> RC
+    RC --> GA
+```
+
+### Plugin Maturity Guidelines
+
+> [!NOTE]
+> New features or sufficiently big code changes to a plugin should lead to:
+> * an increase of the plugin version
+> * knocking the maturity level down.
+
+* First a plugin is Alpha
+* After at least 1.5 months (or more, in 3 month increments, until the quality is deemed acceptable) it becomes Gamma. Meanwhile it can transition through Beta as the maintainer wants (e,g. it can start directly from beta)
+* After at least 3 more months (or more, in 3 month increments — depending on the bug inflow) it becomes Stable
+
+If a plugin must be statically compiled into the server, it is a subject to an additional limitation — its maturity cannot be lower than the server's.
+
+## Beta Plugins In GA Releases
+
+If a plugin contribution is added directly at beta maturity to a GA release series, it will not be removed from that release series. However, we reserve the right to exclude a plugin from future release series if it is unmaintained (for example, if it stops receiving bug fixes).
+
 <sub>_This page is licensed: CC BY-SA / Gnu FDL_</sub>
 
 {% @marketo/form formId="4316" %}
--- a/server/reference/product-development/plugin-development/development-writing-plugins-for-mariadb.md
+++ b/server/reference/product-development/plugin-development/development-writing-plugins-for-mariadb.md
@@ -73,7 +73,7 @@ maria_declare_plugin_end;
 
 ## Maturity Guidelines For Plugins
 
-### Plugin Maturity At a Galnce
+### Plugin Maturity At a Glance
 
 ```mermaid
 stateDiagram-v2


server/reference/product-development/plugin-development/development-writing-plugins-for-mariadb.md
```diff
@@ -71,6 +71,55 @@ maria_declare_plugin(example)
 maria_declare_plugin_end;
 ```
 
+## Maturity Guidelines For Plugins
+
+The maturity level selection criteria differ based on how the plugin is released.
+
+> [!NOTE]
+> Sufficiently big changes to a plugin should lead to:
```
vuvova: "New features or sufficiently big code changes"

server/reference/product-development/plugin-development/development-writing-plugins-for-mariadb.md
```diff
@@ -71,6 +71,55 @@ maria_declare_plugin(example)
...
+> * an increase of the plugin version
+> * knocking the maturity level down.
+
+### Static plugins
+
+Static plugins and other server features share a release cycle.
```
vuvova: "share a [server release cycle](https://mariadb.com/docs/release-notes/community-server/about/release-model)"

vuvova (approved):
I've left a couple of comments. I know Ralf has some too.

mariadb-RalfGebhardt (changes requested):
A plugin with 'experimental' maturity should not be included in a GA release; it should only appear in Previews and RCs. While I understand this might seem to conflict with the strict separation of plugin and server maturity levels—and the fact that the plugin won't be loaded by default—from the server point of view, features must not be removed from a GA Server release. Removing an experimental plugin later would violate that rule.

Whether keeping it is acceptable for an RC is questionable, but I will leave that to other reviewers.

However, this introduces a strict requirement: if a plugin contribution is added directly with 'beta' maturity to a GA release series, it cannot be removed from that release series later. Therefore, we must ensure that the contributor commits to maintaining it.

server/reference/product-development/plugin-development/development-writing-plugins-for-mariadb.md
```diff
@@ -71,6 +71,55 @@ maria_declare_plugin(example)
...
+The maturity level selection criteria differ based on how the plugin is released.
+
+> [!NOTE]
+> New features or sufficiently big code changes to a plugin should lead to:
+> * an increase of the plugin version
+> * knocking the maturity level down.
+
+### Static plugins
+
+Static plugins and other server features share the [server release cycle](https://mariadb.com/docs/release-notes/community-server/about/release-model).
```
vuvova: Sorry. I've just realized that statically linked plugins can also freely use Alpha/Beta as they like.

The hard requirement is — **they cannot have a maturity less than the server**. We can package a dynamic alpha plugin with the GA server. But a static plugin is part of the server binary. And if the server binary is marked stable, every part of it has to be stable.

So, a static plugin can be alpha or beta when the server is still experimental.

gkodinov: I've reworded the whole section a bit.

gkodinov:
I believe I have now resolved @vuvova and @mariadb-RalfGebhardt 's comments. Please check/approve.

server/reference/product-development/plugin-development/development-writing-plugins-for-mariadb.md
vuvova: What about simplifying the whole thing? There's now little difference between static and dynamic, so what about removing separate sections and have, like

----
Plugins released together with the server must follow a unified maturity criteria. For plugins released separately this is a guideline.

* First a plugin is **Alpha**
* After at least 1.5 months (or more, in 3 month increments, until the quality is deemed acceptable) it becomes **Gamma**. Meanwhile it can transition through **Beta** as the maintainer wants (e,g. it can start directly from beta)
* After at least 3 more months (or more, in 3 month increments — depending on the bug inflow) it becomes **Stable**

If a plugin must be statically compiled into the server, it is a subject to an additional limitation — its maturity cannot be lower than the server's.

----
here I merged "static" and "dynamic" into one section. Also removed "experimental" — it doesn't add much value and Ralf said that experimental plugins cannot be added to the GA server, so by skipping experimental we don't need to go into this detail as well.

server/reference/product-development/plugin-development/development-writing-plugins-for-mariadb.md
```diff
@@ -71,6 +71,54 @@ maria_declare_plugin(example)
...
+    RC --> GA
+```
+
+> [!NOTE]
+> New features or sufficiently big code changes to a plugin should lead to:
+> * an increase of the plugin version
+> * knocking the maturity level down.
+
+> [!IMPORTANT]
+> if a plugin contribution is added directly with 'beta' maturity to a GA release series, it **cannot** be removed from that release series later. Therefore, the contributor commits to maintaining it.
```
vuvova: it's not about "how to set the maturity", I'd prefer it to be put elsewhere

vuvova (changes requested):

server/reference/product-development/plugin-development/development-writing-plugins-for-mariadb.md
gkodinov: Serg, dynamic plugins *NOT* released with the server would be too different for that. 
But, frankly, this is not code. So, in the interest of the hope of getting this approved anytime soon, I'm merging.

server/reference/product-development/plugin-development/development-writing-plugins-for-mariadb.md
gkodinov: Any suggestions where "elsewhere" might be? I really would like to avoid guessing.

gkodinov: for now I've just moved it into a new section

server/reference/product-development/plugin-development/development-writing-plugins-for-mariadb.md
```diff
@@ -71,6 +71,42 @@ maria_declare_plugin(example)
...
+* First a plugin is Alpha
+* After at least 1.5 months (or more, in 3 month increments, until the quality is deemed acceptable) it becomes Gamma. Meanwhile it can transition through Beta as the maintainer wants (e,g. it can start directly from beta)
+* After at least 3 more months (or more, in 3 month increments — depending on the bug inflow) it becomes Stable
+
+If a plugin must be statically compiled into the server, it is a subject to an additional limitation — its maturity cannot be lower than the server's.
+
+## Beta Plugins In GA Releases
+
+if a plugin contribution is added directly with 'beta' maturity to a GA release series, it **cannot** be removed from that release series later. Therefore, the contributor commits to maintaining it.
+
```
vuvova, line 109: I'd prefer this section to be removed, it's not about how to transition maturity. It has no real meaning, we cannot force a contributor to maintain a plugin. So I think it's unnecessary and a distraction.

But I'm not going to insist on it, if someone (Ralf?) wants to have it — let's have it.

vuvova (approved):
Looks good. I have a suggestion below — to remove a section, but I'm approving the PR regardless.

server/reference/product-development/plugin-development/development-writing-plugins-for-mariadb.md
mariadb-RalfGebhardt: We can remove this section but need to set the expectations correctly in case an added plugin is not getting bug fixes (is not maintained). Something like that we reserve the right to remove an un-maintained plugin in a next release series.

gkodinov: Thanks, Serg! Yes, indeed it was aiming at addressing Ralf's earlier comment. Let's hope he will have something to say to you about it.

gkodinov: hmm, I'm lost here: I believe the section is setting the explications correctly. So how about just keeping it for now? Note: nothing is set in stone. We can and probably will re-iterate on this again.

mariadb-RalfGebhardt: I recommend to use "If a plugin contribution is added directly at beta maturity to a GA release series, it will not be removed from that release series. However, we reserve the right to exclude a plugin from future release series if it is unmaintained (for example, if it stops receiving bug fixes)."

gkodinov: I've like the text proposed by Ralf. Case closed(?)

PR #583, opened 2026-06-17
Fix errors and inconsistencies for binlog-in-innodb-related docs

knielsen:


ParadoxV5:
🤔 conflicts with [DOCS-5606 (non-public issue)](https://mariadbcorp.atlassian.net/browse/DOCS-5606)

mariadb-stefan-hinz:
Thanks for this — the technical corrections here are valuable and I verified them against the server source (e.g. `innodb_binlog_state_interval` really does default to 2MB not 100MB, `binlog_storage_engine` defaults to none, the removed `SHOW BINLOG EVENTS FROM GTID_POS(...)` example is indeed not valid syntax, GTID-default-since-10.10, `MASTER_DEMOTE_TO_SLAVE`, `replica_pos` alias, etc.).

I pushed one commit on top (`5974718`) with a single purpose: **restore the inclusive primary/replica terminology** in the prose while **keeping all of your technical content and the expanded reference material**. Our docs are mid-migration to primary/replica per [MDEV-18777](https://jira.mariadb.org/browse/MDEV-18777), so we want to keep the prose moving in that direction rather than back to master/slave.

What the commit does:
- Prose flipped back to **primary/replica**; SQL identifiers, keywords, code examples, field names (`Slave_IO_Running`, `Master_Log_File`, …) and option names left **untouched**, and the official REPLICA-form aliases (`SHOW REPLICA STATUS`, `REPLICA MONITOR`, `START/STOP/RESET REPLICA`) used in prose.
- All your technical fixes and the expanded `change-master-to` / `show-replica-status` / `start-replica` / InnoDB-binlog content are retained.
- Fixed two small typos that came in with the PR: `REPLICATIION_REPLICA`, and an incomplete sentence ("Used to transition a primary to become a **.**" → "… to become a replica").

The diff is large only because it touches the same lines you did; net effect is "your content, inclusive wording." Happy to adjust if any of the alias choices read oddly to you. Thanks again for the corrections!

knielsen:
Stefan Hinz ***@***.***> writes:

> reference material**. Our docs are mid-migration to primary/replica
> per [MDEV-18777](https://jira.mariadb.org/browse/MDEV-18777), so we
> want to keep the prose moving in that direction rather than back to
> master/slave.

No, please don't do that, that would be extremely unfortunate. The current
inconsistencies in the documentation are causing a lot of confusion and even
errors, and the problem has become so bad that we urgently need to do
something to solve it, certainly not make it worse.

Everything in the server consistently uses the master/slave terminology. The
documentation needs to describe how the server works, not some fiction that
someone dreamed up.

As the Chief Architect Replication I have the responsibility for the
replication part of the server, you are not making my job any easier this
way. There are plenty of urgent problems to solve in replication with
reliability, performance, and even correctness (and documentation also needs
improving). Let's work together to improve things rather than deliberately
cause new problems.

The second commit in particular urgently needs to be fixed in the
documentation, as this is used to generated the help tables in the source
tree (scripts/fill_help_tables.sql). Until this is done, I will need to
maintain a fork of the documentation with the required consistency fixes,
which is a needless waste of time.

Thanks for your consideration,

 - Kristian.

mariadb-stefan-hinz:
Hi Kristian,
I understand your concerns. However, our documentation guidelines say otherwise. They're stated in this section: https://mariadbcorp.atlassian.net/wiki/spaces/DOCS/pages/3168239632/Documentation+Guidelines#Master-&-Slave
Since that's a page which is available only internally, let me quote that section here:

> Master & Slave
> This is from https://mariadbcorp.atlassian.net/browse/DOCS-5606.
> In all core documentation (Server, MaxScale, ColumnStore, Connectors), change “master” to “primary” and “slave” to “replica”. Notable exceptions:
> Release notes that exist should remain as they are. They reflect the past.
> When parts of the software still use the old terms, the respective core documentation cannot be changed. This is true for
> system variables,
> status variables, and
> information_schema.slave_status table.
> Also, none of the SQL statements have PRIMARY as an alias for MASTER.
> For ColumnStore, use “leader” and “follower”.
> To ensure our documentation is professional, inclusive, and accessible to a global audience, we follow these guidelines for word choice. Our goal is to focus on technical clarity while removing bias and condescension.

It'd be easy enough to change back the terms, but that'd obviously contradict what we've been doing in documentation for well over a year now. I'll bring this up for discussion via email.

ParadoxV5:
I hope this email finds you well, Krisian,

PLC management informed me that the Primary–Replica decision was
discussed last year.
They also said that “The Foundation” has undone your
`CODING_STANDARDS.md` change.
Are you aware of these, or was I misinformed?

Regards,
Jimmy

PR #589, opened 2026-06-18
DOCS-6265 Correct InnoDB-based binary log page

mariadb-stefan-hinz:
Fact-checks and corrects `server/ha-and-performance/standard-replication/innodb-based-binary-log.md` (and the shared `innodb-based-binlog-from-12.3.md` include) against the MariaDB server source — the InnoDB binlog implementation under `storage/innobase/` plus `sql/sys_vars.cc` and `sql/mysqld.cc`.

Filed from [DOCS-6265](https://mariadbcorp.atlassian.net/browse/DOCS-6265), which reported the opening "binlogs are stored in InnoDB tablespaces" claim as wrong. Fact-checking the rest of the page surfaced several more issues.

## Changes

- **Tablespace framing (the reported issue):** binlogs remain separate `.ibb` files on disk; they switch from the flat sequential format to InnoDB's page format, integrated with the redo log and crash recovery. Reworded the frontmatter, intro, and the include accordingly.
- **Wrong variable name:** `binlog_gtid_state` → `gtid_binlog_state` in the migration steps (the registered sysvar is `gtid_binlog_state`).
- **Comparison table:** replaced the incorrect "`.idx`, `.index`" traditional-binlog extension row with an accurate "Files on Disk" row.
- **Removed** the redundant bare `--binlog-storage-engine` enabling stub (full instructions remain below it).
- **Benefit #1** reworded: `--innodb-flush-log-at-trx-commit=0/2` gives binlog↔InnoDB *consistency*, not absolute durability of the latest commits.
- **Added** units/default for `innodb-binlog-state-interval` (bytes, default 2MB, power-of-two multiple of the 16KB page size) and a "Binary Log Encryption" limitation (`--encrypt-binlog` unsupported).
- **Fixed** the broken `mariadb-binlog --read-from-remote-server` example (dangling backslash).
- **Include:** dropped the incorrect "cannot specify a storage location" claim — `binlog_directory` exists for exactly that.

## For reviewer (@brandon-nesterenko)

Two items I could not confirm from the source in my checkout, flagged for technical review:

- **Galera incompatibility** — stated as a design limitation; no explicit code guard found.
- The page still uses the source's own term "InnoDB tablespace files" once (file-structure section). It's source-accurate but is the exact wording the ticket objected to — confirm whether to keep it.

[DOCS-6265]: https://mariadbcorp.atlassian.net/browse/DOCS-6265?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


bnestere (changes requested):
This is a good improvement, thanks @mariadb-stefan-hinz !

One thing to note is that the documentation that Kristian wrote up ([binlog.md](https://github.com/MariaDB/server/blob/main/Docs/replication/binlog.md)) still provides more details in many areas, which would be good to include here too.

Would it make sense to just use that here? Or perhaps Claude could do a comparison between his documentation write-up and this file and fill in the gaps.

mariadb-stefan-hinz:
> This is a good improvement, thanks @mariadb-stefan-hinz !

Thanks! I'll leave it to Uday to review and fold in (the DOCS ticket is assigned to him).

> One thing to note is that the documentation that Kristian wrote up ([binlog.md](https://github.com/MariaDB/server/blob/main/Docs/replication/binlog.md)) still provides more details in many areas, which would be good to include here too.

Wow, that's good to know – and it should have been in the DOCS ticket, but isn't. I've now added it to the ticket.

> Would it make sense to just use that here? Or perhaps Claude could do a comparison between his documentation write-up and this file and fill in the gaps.

I'd vote for the latter. Claude keeps unveiling nasty little errors in draft pages – better to have it fact-check what's in Kristian's page.

mariadb-udaydubey:
@bnestere 
I've pushed a commit that folds in the gaps identified by comparing Kristian's _binlog.md_ with the existing page. Here's a summary of what was added:
Content gaps filled from Kristian's binlog.md:

- Added full Binlog File Format Reference section: file header page offset table and all 7 chunk types.
- Added Third-Party Programs section explaining the stream-based protocol differences for connecting clients
- Added missing limitations: MASTER_POS_WAIT(), BINLOG_GTID_POS(), RESET MASTER TO, --init-rpl-role, savepoints inside triggers (MDEV-38465), --binlog-checksum removed, binlog_gtid_index_hit/binlog_gtid_index_miss not used
- Added XA Transaction Replication limitation — external XA is supported and crash-safe on the master, but replicating XA PREPARE to a replica is not supported
- Added FLUSH BINARY LOGS DELETE_DOMAIN_ID behavior difference from the old implementation
- Added pre-allocation cycle detail (file N → N+1 data, N+2 pre-allocated in background)

Flagging a couple points for review:
- Kristian's binlog.md states only: "The new binlog implementation cannot be used with Galera." with not much explanation. As of now, added as a one-liner. If you could clarify more on this, before the PR is merged.

- Added a new XA Transaction Replication limitation section

Let me know if you need any changes.

server/ha-and-performance/standard-replication/innodb-based-binary-log.md
```diff
@@ -69,7 +65,7 @@ log_bin
 binlog_storage_engine=innodb
 ```
 
-* The `log_bin` option must be specified without any value.
+* The `log_bin` option must be specified without any value; it is not an on/off switch.
```
bnestere: > ; it is not an on/off switch.

I realize this comes from binlog.md, but I think this clause adds more confusion than omitting it. I think

> The `log_bin` option must be specified without any value

suffices. @knielsen what do you think?

server/ha-and-performance/standard-replication/innodb-based-binary-log.md
```diff
@@ -152,17 +153,20 @@ The `mariadb-binlog` utility allows inspection of binary log files either locall
...
 
-# Read multiple binary log files
+# Read multiple binary log files (pass all at once, in order)
 mariadb-binlog /path/to/binlog-000000.ibb \
                /path/to/binlog-000001.ibb \
                /path/to/binlog-000002.ibb
 ```
 
-When viewing events across multiple binlog files, all binlog files should be passed to the `mariadb-binlog` program at once in correct order; this ensures that events that cross file boundaries are included in the output exactly once.
+When viewing events across multiple binlog files, all binlog files should be passed to the `mariadb-binlog` at once in the correct order; this ensures that events that cross file boundaries are included in the output exactly once.
```
bnestere: program was deleted. This should be `mariadb-binlog` **program**

server/ha-and-performance/standard-replication/innodb-based-binary-log.md
```diff
@@ -246,113 +250,199 @@ START SLAVE;
...
 
 ### Migration from Traditional Binlog
 
-Starting with MariaDB 12.3, there are three primary methods to migrate from the Traditional Binlog to the New Binlog Format:
+When switching an existing server to the new binlog format, the old binlog files will not be available after the switch, as the two formats are mutually exclusive. Choose the method that best suits your environment.
 
 #### Method 1: Direct Restart Migration (Simple)
 
-This is the simplest approach, appropriate for cases where point-in-time recovery does not require maintaining old binary log files.
+This is the simplest approach, suitable when maintaining old binary log files for point-in-time recovery is unnecessary.
```
bnestere: Describing the method this way is confusing. You don't describe what "Direct Restart Migration" is, but you provide clarifying details on it from the start. Please first describe what the method _is_, then clarify for what types of setups/configurations this is applicable for/to

server/ha-and-performance/standard-replication/innodb-based-binary-log.md
```diff
@@ -246,113 +250,199 @@ START SLAVE;
...
 
 1. Stop the MariaDB server.
 2. &#x20;Add `--log-bin` and `--binlog-storage-engine=innodb` to your configuration.
```
bnestere: If one is migrating from a tradition binlog, they'd already have `--log-bin`, but possibly with a name provided. Simply adding `--log-bin` isn't correct, they should amend their existing configuration for the variable. The binlog.md doesn't suggest this.

server/ha-and-performance/standard-replication/innodb-based-binary-log.md
```diff
@@ -246,113 +250,199 @@ START SLAVE;
...
 #### Galera Cluster
 
-The InnoDB-based binary log cannot be used with MariaDB Galera Cluster.
+The InnoDB-based binary log cannot be used with MariaDB Galera Cluster. Users running Galera must continue using the traditional binlog format.
+
+#### XA Transaction Replication
+
+External XA transactions are fully supported and crash-safe on the master. The new binlog stores `XA PREPARE` and `XA COMMIT`/`XA ROLLBACK` records internally (see chunk types 5 and 6 in the Binlog File Format Reference), and the server will always recover prepared transactions to a state consistent with the binlog after a crash.
+
+However, replicating `XA PREPARE` to a replica is not supported. This means that if the master has transactions in a prepared-but-not-yet-committed state at the time of a failover, it is not possible to promote the replica and rescue those prepared transactions. In general, this is unlikely to cause issues, as external XA users typically do not rely on this failover path.
```
bnestere: I'd remove

> In general, this is unlikely to cause issues, as external XA users typically do not rely on this failover path.

b/c it is an opinion (and not for us to say). Let's just stick to the facts.

server/ha-and-performance/standard-replication/innodb-based-binary-log.md
```diff
@@ -246,113 +250,199 @@ START SLAVE;
...
+
+#### Binary Log Encryption
+
+Binary log encryption (`--encrypt-binlog`) is not supported with the InnoDB-based binary log. Use full-disk encryption at the operating-system level instead, and/or SSL for the replica's connection to the master (use the `MASTER_SSL` option in `CHANGE MASTER`).
```
bnestere: @knielsen I thought we could hook into innodb's encryption here. Is that not yet implemented?

bnestere (changes requested):
@mariadb-udaydubey / @mariadb-stefan-hinz I've left a few comments, though I also would like some input from @knielsen for some of the points.

mariadb-udaydubey:
@bnestere 
I have pushed an updated commit addressing the feedback. Here's a summary of what was changed and what is still pending:

- Restructured the opening line to first describe what the method does, then clarify what types of setups and configurations it is applicable for
- Followed the same for other two methods for consistency (Replication State Migration (GTID Preservation and Live Migration (Zero Downtime))
- Added "mariadb-binlog program" wording
- Removed "; it is not an on/off switch" clause as suggested (as of now and we can restore if knielsen suggest)
- Removed the opinion sentence "In practice this is unlikely to cause issues, as external XA users typically do not rely on this failover path"

Pending Engineering Input (@knielsen)
Two items are left unchanged pending Kristian's response:
- Binary Log Encryption
- Galera Cluster

Let me know if you need any further changes.

mariadb-udaydubey:
Made some minor improvements:
- Changed "applicable for" to "applies to" for clarity
- Added blank line before list items for proper Markdown rendering

bnestere (approved):
Thanks @mariadb-udaydubey !

Looks good. Feel free to merge this in, and we can create a separate PR for any other follow ups.

PR #603, opened 2026-06-21
These are C/ODBC 3.1.23 and 3.2.9 release notes

lawrinn:
Things have gone wrong and I can't fix it as I can't make a usable local clone(not sure why) and I can't even re-base. So, here 2 release notes together.


lawrinn:
i'm re-doing it in 2 separate PR's

PR #609, opened 2026-06-22
Track in-flight skill proposals in /weekly-review + fix transcript path

mariadb-stefan-hinz:
Two changes to `.claude/commands/weekly-review.md`:

1. **Track in-flight proposals.** Adds an inputs item (scan `dev-docs/skill-proposals/*.md` and `dev-docs/agent-improvements/*.md`) and a new output section reporting each open proposal's status + dogfood signal + recommended next step — so proposals like `jira-chase` (DOCS-6274) don't stall between filing and promotion.
2. **Fix the transcript-scan path.** It was hardcoded to a teammate's project dir (`-Users-tauseefkhanmdb-Documents-GitHub-mariadb-docs`), so the usage-summary scan silently matched nothing on other machines. Repointed to `-Users-stefanhinz-maria` with an inline note that the path is machine-specific and should be adjusted per checkout.

Analysis-only command; no behavior beyond reporting.

### DIFF
--- a/.claude/commands/weekly-review.md
+++ b/.claude/commands/weekly-review.md
@@ -9,11 +9,11 @@ report at the end.
 
 ## Inputs (all LOCAL — that's why this runs locally, not as a cloud routine)
 
-1. **Session transcripts** — under `~/.claude/projects/-Users-stefanhinz-maria/`.
-   (Machine-specific: this is the project dir for a session launched from `/Users/stefanhinz/maria`;
-   adjust it to your own checkout's project dir if you run this elsewhere.)
-   Find session files modified in the last 7 days:
-   `find ~/.claude/projects/-Users-stefanhinz-maria -name '*.jsonl' -mtime -7 -print` (also check `.json`).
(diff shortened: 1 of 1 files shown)


mariadb-stefan-hinz:
Replaced the transcript-path swap (commit e29fb2df2).

The original change repointed the hardcoded path from `-Users-tauseefkhanmdb-Documents-GitHub-mariadb-docs` to `-Users-stefanhinz-maria`. But that path is the **Claude Code project-dir name**, which is mechanically derived from the directory Claude was launched in (path separators and `.` → `-`) — so it's inherently per-machine. Hardcoding it to either of us just moves the breakage to the other (and to anyone on a third machine).

So instead of swapping whose path is baked in, it now derives the dir at runtime from `pwd`:

```bash
find ~/.claude/projects/"$(pwd | sed 's#[/.]#-#g')" -name '*.jsonl' -mtime -7 -print
```

Nothing to maintain, works on any checkout. The "track in-flight proposals" change is untouched.

PR #620, opened 2026-06-24
DOCS-6276 Fix MYSQL_BIND length/error/is_null member descriptions

mariadb-udaydubey:
As per the ticket description, the _MYSQL_BIND_ table claimed length_value, error_value, and is_null_value are used when the corresponding length/error/is_null pointers are NULL. Connector/C does not use them automatically; they are only used when their address is manually assigned to the pointer.


mariadb-stefan-hinz (approved):
Verified the fix against the Connector/C source (local clone of mariadb-corporation/mariadb-connector-c).

The change is correct. In `mysql_stmt_bind_result()` (`libmariadb/mariadb_stmt.c:1423-1439`), the connector copies the caller's bind array into its **internal** `stmt->bind` and then, when a `length`/`is_null`/`error` pointer is left NULL, redirects the *internal* copy's pointer to its *internal* `*_value` member:

```c
memcpy(stmt->bind, bind, sizeof(MYSQL_BIND) * stmt->field_count);
...
if (!stmt->bind[i].length)
  stmt->bind[i].length = &stmt->bind[i].length_value;
```

So fetch results land in the connector's copy, not the caller's struct. The caller's `length_value`/`error_value`/`is_null_value` are populated only if the caller assigns their address to the corresponding pointer — exactly the new wording. The struct's own header comments ("Used if length is 0", `include/mariadb_stmt.h:130`) match the old text and are what made the page misleading. LGTM.

Two pre-existing issues in the same `MYSQL_BIND` table, not introduced by this PR but worth fixing while it's open (non-blocking):

1. **Duplicate `is_null_value` row** — the table lists `is_null_value` twice; the last row (`Used if value is NULL`) is a leftover and should be removed (the struct has a single `is_null_value` member).
2. **Typo** — the `is_null` pointer row spells the member `is_nulll` (three l's).

Approving; happy to re-look if you fold the two nits in.

PR #644, opened 2026-06-29
Add missing link in docs

BrianDouglasIE:
Hi,

The "Getting Started With the Node.js Connector" documentation page has a sentence that seems to be incomplete. I've added what I presume is the missing link to the Typescript docs. I've also slightly altered the sentence for readability.

Page in question: https://mariadb.com/docs/connectors/mariadb-connector-nodejs/getting-started-with-the-node-js-connector

You can let me know if this is helpful or not.

Thanks,

Brian


mariadb-stefan-hinz:
Thanks for the fix! I made the link relative – otherwise, looks good!

PR #645, opened 2026-06-29
MXS Update MaxScale Trial documentation

jhnwkmn:
This should be merged only when MaxScale 25.10.3 Trial has been published.


mariadb-stefan-hinz:
LGTM. Merged.

PR #689, opened 2026-07-03
Fix escapes in string-literals.md

ParadoxV5:
Markdown also uses `\` for escaping.


mariadb-stefan-hinz:
That doesn't look right. Sure, some entries were missing that didn't survive the KB-to-GitBook migration – thanks for adding them! But escaping doesn't necessarily mean double backslashes – for example, \0 comes out correctly on the published page (https://mariadb.com/docs/server/reference/sql-structure/sql-language-structure/string-literals#escape-sequences).
I'm cancelling this PR, but will make sure the *missing* escape sequences are added to that page.

PR #740, opened 2026-07-14
DOCS-5089 Document NBO mode limitations on Galera schema-upgrade page

mariadb-udaydubey:
List NBO's supported and unsupported DDL statements explicitly (adding REPAIR and LOCK = NONE), and add a warning that running an unsupported statement under wsrep_OSU_method = 'NBO' returns an error rather than falling back, so NBO should be set per-session, not server-wide.

### DIFF
--- a/galera-cluster/galera-management/general-operations/performing-schema-upgrades-in-galera-cluster.md
+++ b/galera-cluster/galera-management/general-operations/performing-schema-upgrades-in-galera-cluster.md
@@ -145,11 +145,16 @@ Like the TOI method, NBO replicates DDL statements to all nodes in the cluster s
 
 ### **Key Considerations for NBO Syntax**
 
-To utilize NBO, the DDL statement must meet specific criteria:
+NBO supports only a specific set of DDL statements. To use NBO, the statement must meet these criteria:
 
-* Explicit Locking: `ALTER TABLE` statements must include an explicit `LOCK SHARED` or `LOCK EXCLUSIVE` clause. Statements without a `LOCK` clause default to `DEFAULT` and are not supported.
-* Supported Commands: Beyond specific `ALTER` statements, `ANALYZE TABLE` and `OPTIMIZE TABLE` are also supported.
-* Single Table Limitation: Do not use NBO with statements that operate on more than one table at a time.
+* **Explicit locking:** `ALTER TABLE` statements must include an explicit `LOCK = SHARED` or `LOCK = EXCLUSIVE` clause. A statement with no `LOCK` clause, or with `LOCK = DEFAULT` or `LOCK = NONE`, is **not** supported.
+* **Supported statements:** `ALTER TABLE ... LOCK = {SHARED | EXCLUSIVE}` (including the partition-management form), `ANALYZE TABLE`, and `OPTIMIZE TABLE`.
+* **Unsupported statements:** `CREATE`, `RENAME`, `DROP`, and `REPAIR`.
+* **Single-table only:** do not use NBO with statements that operate on more than one table.
+
+{% hint style="warning" %}
+Running an unsupported statement while `wsrep_OSU_method = 'NBO'` **returns an error** rather than falling back to another method. For this reason, set NBO at the session level for the specific statements that support it — never server-wide — and run `CREATE`, `DROP`, and similar statements under the default `TOI` method.
+{% endhint %}
 
 ### **When to Use NBO**
 


galera-cluster/galera-management/general-operations/performing-schema-upgrades-in-galera-cluster.md
```diff
@@ -145,11 +145,16 @@ Like the TOI method, NBO replicates DDL statements to all nodes in the cluster s
 
 ### **Key Considerations for NBO Syntax**
 
-To utilize NBO, the DDL statement must meet specific criteria:
+NBO supports only a specific set of DDL statements. To use NBO, the statement must meet these criteria:
 
-* Explicit Locking: `ALTER TABLE` statements must include an explicit `LOCK SHARED` or `LOCK EXCLUSIVE` clause. Statements without a `LOCK` clause default to `DEFAULT` and are not supported.
-* Supported Commands: Beyond specific `ALTER` statements, `ANALYZE TABLE` and `OPTIMIZE TABLE` are also supported.
-* Single Table Limitation: Do not use NBO with statements that operate on more than one table at a time.
+* **Explicit locking:** `ALTER TABLE` statements must include an explicit `LOCK = SHARED` or `LOCK = EXCLUSIVE` clause. A statement with no `LOCK` clause, or with `LOCK = DEFAULT` or `LOCK = NONE`, is **not** supported.
```
mariadb-TeemuOllakka, line 150: Hi, I verified these from code:

* `DROP TABLE` has been supported since ES 10.6.18-14
* `CREATE INDEX` is supported if it has explicit `LOCK` clause, just as `ALTER TABLE`
* `DROP INDEX` is supported
* Single table rule is checked explicitly only for `OPTIMIZE TABLE`
* `ANALYZE TABLE` is not supported

mariadb-TeemuOllakka (changes requested):

galera-cluster/galera-management/general-operations/performing-schema-upgrades-in-galera-cluster.md
mariadb-udaydubey: @mariadb-TeemuOllakka 
Thanks Teemu. Really appreciate the code-level check. I confirmed on my end that NBO isn't in community source at all (no OSU enum value, no test coverage, wsrep-lib not populated), so I can't independently verify these against ES source myself, but I trust your read here.

Before I fix this, I want to ensure that the version-specific details are clearly stated and accurate, rather than being left to interpretation. For example, you provided the minimum supported version for DROP TABLE (ES 10.6.18-14). To document the rest correctly:

1. Do `CREATE INDEX` and `DROP INDEX` have their own version floors, or have they always been supported alongside ALTER TABLE?
2. For `ANALYZE TABLE` — was it never supported, or was support removed at some point? (Our page currently says it is supported, so I want to phrase the correction accurately either way.)
3. Does the `DROP TABLE` floor of 10.6.18-14 have an equivalent version on the 11.4 series, since this page serves both release lines?

Once I have those, I'll fix this bullet plus the same claims repeated in a few other spots on the page (Disadvantages section, the warning callout, and "When to Use") so it's consistent throughout and push an updated commit.

mariadb-udaydubey: Thanks Teemu — since this sat for a while with no reply to my follow-up questions, I went ahead and applied your five corrections as given, without inventing version floors beyond the one you supplied:

- `DROP TABLE` supported (since MariaDB Enterprise Server 10.6.18-14, as you stated)
- `CREATE INDEX` supported with an explicit `LOCK` clause, same as `ALTER TABLE`
- `DROP INDEX` supported
- Single-table check enforced explicitly only for `OPTIMIZE TABLE`
- `ANALYZE TABLE` **not** supported

Applied consistently across the Disadvantages, Key Considerations, warning callout, and When to Use sections (`7b4ab7143`).

I also independently re-verified all five against the ES source itself (`wsrep_can_run_in_nbo()` in `sql/wsrep_mysqld.cc`) rather than relying on the review alone — everything checks out exactly as you described. The `DROP TABLE` floor traces to commit `6068418a7e3` ("MENT-2074: Add support for DROP TABLE in NBO mode"), and `10.6.18-14` is exactly the first 10.6.x release branch containing it — matches your number precisely.

Thanks again for the code-level check, really appreciate it. Ready for another look whenever you have time.

PR #764, opened 2026-07-20
Fix Rows Event errors

ParadoxV5:
* Fix Hexadecimal Event IDs for `WRITE_ROWS_EVENT` and `UPDATE_ROWS_EVENT`
* Fix “Number of columns” missing interger length


mariadb-stefan-hinz:
Thanks for the fix!

PR #827, opened 2026-08-03
MXS-6766: Add security vulnerability list for MaxScale

markus456:
The list currently only contains CVE-2023-40354 which was added in the ticket.


mariadb-AllenHerrera:
Looks good

PR #834, opened 2026-08-04
MXS Update release notes scripts

jhnwkmn:


dbart (commented):
Reviewed the scripts and cross-checked the generated output against the published MaxScale release notes. The `md_escape` idea is a real improvement and I've kept it — but it had a corruption bug, and two things in this branch went stale because `main` moved after you cut it.

**I've pushed the fixes to this branch** (`cd9a64c71`) rather than leaving you a list. One item below still needs your call, and I deliberately left the nits alone. Details so you can check my work.

---

## Main moved under you — two of these aren't your mistakes

**1. The license-notice include was correct when you branched.** At this PR's base (`d51495f474`), `release-notes/maxscale/25.10/25.10.3.md` still ended with:

```
{% include "https://app.gitbook.com/s/SsmexDFPv2xG2OTyO5yV/~/reusable/pNHZQXPP5OEz2TgvhFva/" %}
```

So adding it to the generator matched the convention as it stood. The `DOCS-6394 Replace licence includes with literal notices` batches 1–6 then removed that include repo-wide; `main` now has zero occurrences under `release-notes/`, and every MaxScale series from 22.08 onward carries the notice inline instead:

```
<sub>_This page is: Copyright © 2026 MariaDB. All rights reserved._</sub>
```

Fixed by emitting the literal notice. The generated tail is now byte-identical to `main`'s `25.10/25.10.3.md`. I defined it in a variable above the heredoc so the rationale comment doesn't get emitted into the generated Markdown, and noted there that the year is hardcoded on purpose — it keeps generated pages agreeing with their siblings, and wants the same one-line bump whenever the repo-wide notice year moves.

**2. The `mariadb-cloud-byoa-provisioning.md` change was never yours.** It came from bot commit `8449d7db5 docs: expand GitBook aliases`, running the rule as it existed at your base:

```
-e 's|{mariadb-cloud}|https://app.gitbook.com/o/.../s/vPz15Lz0Iw3P3yKR3Prd|g' \
```

Unguarded, so it expanded `{mariadb-cloud}` in **bare prose inside an HTML comment** — mangling a note that told a human to convert links *back to* aliases into "convert the cross-links back to `https://app.gitbook.com/...` aliases". `fa33cfd12 DOCS-6481 Expand GitBook aliases in link targets only` has since fixed the workflow, and `main` has rewritten that comment block three times, so the branch had stopped merging cleanly:

```
$ git merge-tree --write-tree pr834 origin/main
CONFLICT (content): Merge conflict in .../mariadb-cloud-byoa-provisioning.md
```

Fixed by restoring the file to `main`'s content. Verified `NO CONFLICTS` against `origin/main` on the pushed tip. The hardened workflow won't reintroduce it.

---

## Real bug in `md_escape`

Backslashes are **literal inside a code span**, so escaping there emits a visible `\`. MXS-5837 is in the corpus three times (`24.02/24.02.7.md:29`, `22.08/22.08.18.md:26`, `23.02/23.02.15.md:27`) and would have regressed:

```
before: (`std::length_error` in `maxbase::load_file<std::string>(...)`)
after:  (`std::length\_error` in `maxbase::load\_file<std::string>(...)`)   ← visible backslashes
```

Also extended the character set. `_` alone isn't enough — `[` and `*` both occur in real MXS summaries and GitBook's own normalizer escapes them, e.g. `\[readwritesplit]`, `\[Maintenance]`, `admin\_ssl\_\*`. Matching that keeps generated output byte-identical to what GitBook would save, so the first edit in the web app doesn't produce a spurious `GITBOOK-XXX` diff.

The replacement is code-span aware: split on backticks, escape only the even-index parts. Tested against seven strings copied verbatim out of published pages, 7/7 match — including the code-span cases, which now pass through untouched.

---

## Also added

Recent point releases carry a **Release Date** line (`25.10.1`–`25.10.3`, `25.01.6`–`25.01.7`); the template omitted it. Added as an optional third argument:

```
usage: generate_release_notes.sh major.minor.patch [maturity] [release-date]
```

Omit it and you get a `TBD` placeholder plus a loud warning after the manual-update list, so it can't ship unnoticed.

---

## Needs your call

**The H1 dropping "MariaDB".** `dev-docs/style-guide.md:85` says the product name is **MariaDB MaxScale**, but actual practice is genuinely split — 76 `# MaxScale …` vs 62 `# MariaDB MaxScale …` across the MaxScale release notes, and recent point releases do favour your version (`25.10.0`, `25.10.1`, `25.10.3`), while the series READMEs keep "MariaDB MaxScale". I've **left your change as-is** since it matches recent practice for exactly the pages this script generates. But it shouldn't become house style silently via a generator — either revert it here, or land a style-guide update alongside so the next person doesn't "fix" it back.

---

## Left alone deliberately

- **Raw `app.gitbook.com` URLs instead of `{maxscale}` aliases.** Golden rule 3 would prefer aliases, but the Action only expands on PR branches — a release note committed straight to `main` would ship a literal `{maxscale}`. The hardcoded URLs work; not worth the failure mode. Worth knowing they also lack the `/o/diTpXxF5WsbHqTReoBsS` org segment the Action now emits.
- **`upgrade_version` (line 41) is dead code** — computed, never used, and its comment implies maintenance that does nothing. Unrelated churn for this PR.

## Verified fine, don't "fix" these

- `{% @marketo/form formid="4316" formId="4316" %}` — the duplicated attribute looks like a typo but is the house form (2377 occurrences vs 7 without). Correct as written.
- **Unwrapping the long lines** is a positive: GitBook reflows to long lines on save, so pre-unwrapped output means less round-trip churn.

Checks on the pushed tip: `bash -n` and `py_compile` clean, `doc-lint.sh` exit 0, codespell exit 0, no merge conflicts against `main`. Note the branch is ~3 weeks behind `main` — worth a rebase before merge, though nothing conflicts.

One process nit: no DOCS ticket in the commit subject or PR body (repo convention is `DOCS-XXXX`).

PR #839, opened 2026-08-05
DOCS-6395 Fix stale EXPLAIN FORMAT=JSON example and column name

mariadb-udaydubey:
The Synopsis example predated the nested_loop wrapper and the loops/cost members, and wrongly parenthesized attached_condition. The page also said the output column is titled "JSON" while its own example showed EXPLAIN.

Replace the example with verbatim recorded output from mysql-test/main/sargable_casefold.result:160-177, which has real unmasked cost values, together with the CREATE TABLE it was measured against so the figures correspond to a reproducible schema. Correct the column name to EXPLAIN, and note that ANALYZE FORMAT=JSON titles it ANALYZE.

Verified against MariaDB server @ c0b303f8741d6e1cdeb871618617f6a557004e01.


dbart (commented):
Reviewed this against a live server rather than only against source, since the whole point of the change is that the recorded example should match what a reader actually sees.

**Server:** `11.8.8-5-MariaDB-enterprise`. I rebuilt the schema and data exactly as `mysql-test/main/sargable_casefold.result:4-19` does (same `CREATE TABLE`, same `seq_1_to_100` INSERT) and ran the examples.

## Everything structural checks out

| Claim | Live result |
| --- | --- |
| Column titled `EXPLAIN`, not `JSON` | Confirmed — raw header is `EXPLAIN` |
| `ANALYZE FORMAT=JSON` titles it `ANALYZE` | Confirmed — `ANALYZE: {` |
| `nested_loop` wrapper always present | Confirmed |
| Message-only plans stay un-nested | Confirmed — `"table": {"message": "Impossible WHERE"}`, no `nested_loop` |
| `loops` and `cost` at both levels | Confirmed |
| `sargable_casefold` on by default | Confirmed — `UPPER()` stripped, condition is `t1.col2 = t1.col1` |

Also worth recording: I settled the `attached_condition` contradiction empirically. Running the **old** page's own query, `WHERE col1=1`, gives:

```
"attached_condition": "t1.col1 = 1"
```

No parentheses. The ticket **description** was right, the 2026-08-03 comment was wrong, and the "Correction note on claim 7" in the fact-check report resolved it the correct way. That's now confirmed on a running server, not just from `.result` files.

I also checked the recorded block on the **11.8** branch, not just the `main`/13.1 SHA the report pins — `sargable_casefold.result:160-177` is byte-identical there, cost values included. So the example isn't main-only, which is good given the page covers the older series too.

One non-issue, in case it comes up: `ANALYZE TABLE ... PERSISTENT FOR ALL` (`:20` in the test) makes **no** difference to this plan. Cost was identical before and after it. No need to add it to the page.

## One thing to fix: the cost caveat

The cost value doesn't reproduce:

```
PR / sargable_casefold.result (11.8 and main):   "cost": 0.0256761
Live 11.8.8-5-MariaDB-enterprise:                "cost": 0.0271548
```

Identical schema, identical data, same major version — and it still differs, at both `query_block` and table level. The variable here is the server build/configuration.

That's not a problem with the example itself; it's a problem with how the page explains it:

> Cost values depend on the data and the hardware, so the exact figures vary between servers.

It names *data* and *hardware*, and here neither is the cause — the data was byte-identical. A reader who reproduces the schema faithfully and gets a different number has been told the two things that could explain it, and both are ruled out. Suggested reword:

> Cost values depend on the data, the server build, and the hardware, so the exact figures vary between servers.

(Adding the test's `INSERT ... FROM seq_1_to_100` to the page is optional — it would help a reader match `rows: 100`, but as shown above it won't make the cost match, so it doesn't fix the caveat on its own.)

## Minor editorial nit

> Table access is always nested inside a `nested_loop` array. The exception is a plan that only carries a message...

`always` and `exception` fight each other. Suggest:

> Table access is nested inside a `nested_loop` array, except in a plan that only carries a message, such as `{"table": {"message": "Impossible WHERE"}}`.

## CI gates — I ran the two that were skipped locally

The reviewer note flagged that `codespell` and `lychee` weren't available on the authoring machine. Both are installed here, so I ran `doc-lint.sh` against this branch:

- **codespell — PASS**
- **lychee — FAIL**, on exactly the link the note predicted:

```
[ERROR] https://s.petrunia.net/blog/?p=93 (at 71:117)
        Connection refused - server may be down or port blocked
```

Confirmed **pre-existing and not a regression** — the link is untouched by this diff and already lives on `main`. But it is genuinely dead, not a transient blip: DNS resolves (`90.156.144.163`) while both :80 and :443 refuse connections. Since lychee gates the whole changed file, **this will fail CI on this PR** even though the PR didn't cause it.

There's a working Wayback capture of the same post (verified — it's the right one, MySQL Workbench / Visual Explain / `EXPLAIN FORMAT=JSON` parsing):

```
https://web.archive.org/web/20200218115814/http://s.petrunia.net:80/blog/?p=93
```

Swapping the URL in that line would unblock CI. Happy for it to go in here as a drive-by, or as a separate ticket if you'd rather keep this PR to its stated scope — your call, but it does need to happen for this to go green.

## Verdict

Technically sound, and the underlying research is solid — everything the PR set out to fix is genuinely broken on the published page and genuinely fixed here. Approve once the cost caveat is reworded and the dead link is dealt with.

mariadb-udaydubey:
@dbart thanks for the thorough live-server review — all three of your points are addressed in 212b9027f:

1. **Cost caveat** reworded to name the server build as a factor: *"Cost values depend on the data, the server build, and the hardware, so the exact figures vary between servers."* — matches your empirical finding that byte-identical data on the same major version still produced a different cost.
2. **`nested_loop` wording** fixed — the "always … The exception" contradiction is gone: *"Table access is nested inside a `nested_loop` array, except in a plan that only carries a message …"*
3. **Dead `s.petrunia.net` link** swapped for your verified Wayback capture (`web.archive.org/web/20200218115814/…`).

One note on the CI gate: `s.petrunia.net` is already on the lychee exclude list in `link-check-pr.yml`, so `link-check` was passing on this PR regardless — the dead link wasn't actually blocking CI here. I swapped it anyway since it's genuinely dead for readers, and `web.archive.org` is excluded on `main`, so CI stays green. Could you re-review and approve if it looks good?

dbart (approved):
Approved!

PR #843, opened 2026-08-05
DOCS-6396 Correct r_rows values in ANALYZE examples

mariadb-udaydubey:
All three ANALYZE output examples on the page showed r_rows as a bare integer. The server builds that column as a two-decimal float -- Item_float(thd, avg_rows, 2) at sql/sql_explain.cc:1653 -- so it is never printed without decimals. Recorded output in mysql-test/main/ analyze_stmt.result:10 confirms: r_rows reads 10.00, not 10. Verified against server origin/main @ 87bcaaaa.

Fix r_rows in the vertical example (181.00) and in both tabular examples (150000.00, 10.00), widening the column from 6 to 9 characters. The NULL in the second example is correct as-is and stays: when a table is never scanned, sql_explain.cc:1646-1649 pushes item_null and bypasses the Item_float path entirely. Update the prose at L85 and L87, which quoted the same wrong values.

Also drop the editorial ellipsis from possible_keys. Real EXPLAIN never emits "..." -- push_string_list() at sql/sql_explain.cc:382-398 joins the key list with a bare comma and has no truncation -- so PRIMARY,... is not output the server can produce. The page gives no CREATE TABLE, so the true index list is unknowable; plain PRIMARY is used rather than inventing index names.

Fix a typo in the second example query: c_acctbal < -0 becomes < 0, matching the same query at L70 and the prose at L113.

One known deviation is left in place deliberately. The ref column is database-qualified in real output (dbt3_s001.orders.o_orderkey in mysql-test/main/rowid_filter.result:331), but the example's 150000-row customer table is dbt3 scale factor 1 while the test database dbt3_s001 is scale 0.01, so that qualifier would contradict the numbers shown. No database name is derivable from the page, so ref is left unqualified rather than invented.

### DIFF
--- a/server/reference/sql-statements/administrative-sql-statements/analyze-and-explain-statements/analyze-statement.md
+++ b/server/reference/sql-statements/administrative-sql-statements/analyze-and-explain-statements/analyze-statement.md
@@ -43,7 +43,7 @@ possible_keys: key1
       key_len: 5
           ref: NULL
          rows: 181
-       r_rows: 181
+       r_rows: 181.00
      filtered: 100.00
    r_filtered: 10.50
         Extra: Using index condition; Using where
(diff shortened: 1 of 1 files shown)


dbart:
Pushed 113c9449 to this branch — a two-line follow-on to the `r_rows` fix. Flagging it here since it's your ticket.

### What changed

The `orders` row in both tabular examples now left-aligns `rows` and `r_rows`:

```
before:  | customer.c_custkey |      7 |     10.00 |   100.00 |      30.03 |
after:   | customer.c_custkey | 7      | 10.00     |   100.00 |      30.03 |
```

### Why

The client right-justifies a column only when `IS_NUM(field->type)` holds — `num_flag[…] = IS_NUM(field->type)` at `client/mysql.cc:4029`, applied at `:4101`. `rows` and `r_rows` are both built as `Item_string_sys` and arrive as `VAR_STRING`, so real output left-aligns them; only `filtered` and `r_filtered`, which are `DOUBLE`, are right-aligned. `--column-type-info` confirms the types.

Verified on a live 11.8.8-5-MariaDB-enterprise server, with dbt3-shaped tables (150000 customers, 1.5M orders) and histograms collected so the optimizer picked the same plan the page shows:

```
| id   | select_type | table    | type | possible_keys | key         | key_len | ref                         | rows   | r_rows    | filtered | r_filtered | Extra       |
|    1 | SIMPLE      | customer | ALL  | PRIMARY       | NULL        | NULL    | NULL                        | 150000 | 150000.00 |    18.00 |      18.00 | Using where |
|    1 | SIMPLE      | orders   | ref  | i_o_custkey   | i_o_custkey | 5       | docs6396.customer.c_custkey | 10     | 10.00     |    30.00 |      33.33 | Using where |
```

Only the `orders` rows needed changing — the `customer` row's values (`149095`, `150000.00`) fill their columns exactly, so the padding isn't visible there. Column widths are unchanged, and `key_len` (also `VAR_STRING`) was already correctly left-aligned.

### The rest of the PR checks out on live output

Recording this because the report could only argue some of it from source:

- Example 1 reproduces field-for-field (modulo table name), `r_rows: 181.00` included — `rows: 181`, `filtered: 100.00`, `r_filtered: 10.50`, `key_len: 5`, and `Extra: Using index condition; Using where` all match.
- `150000.00` and `10.00` are exactly what the server prints for the join example, and example 3 gives bare `NULL` for `r_rows` with `r_filtered` NULL alongside it.
- **`possible_keys: PRIMARY` is a stronger call than the report claims.** It's not merely "a form the server can emit" — it is precisely what this plan emits.
- `< -0` and `< 0` produce identical output, so that change is safe.
- The `ref` deviation is real (live: `docs6396.customer.c_custkey`), and leaving it unqualified was defensible. One narrowing, though: only `dbt3_s001` specifically conflicts with the row counts. A neutral `dbt3.customer.c_custkey` would be consistent with the numbers shown and would close the deviation, if you'd rather not leave it open.

### Two small corrections to the write-up

"Real EXPLAIN never emits `...`" is a little too strong — EXPLAIN *can* emit `...` in the `table` column for a UNION with more members than `NAME_LEN` allows (`sql/sql_explain.cc:570`, `"...>"`). It holds for `possible_keys`, where `push_string_list` joins with a bare comma and has no truncation, so the fix itself is right; just the rationale overreaches.

And for whoever revisits the JSON pages later: `ANALYZE FORMAT=JSON` emits `r_rows` via `add_double` (minimal representation), not `Item_float(…, 2)`, so `"r_rows": 150000` and `"r_rows": 10` on `analyze-formatjson-examples.md` and friends are **correct** and shouldn't be "fixed" to match. Scoping this ticket to the tabular form was right — that reasoning just isn't recorded anywhere yet.

Filing a separate follow-up for `agent-skills/granular/statements/mariadb-explain/SKILL.md:127`, which carries this same example with `r_rows: 181`.

dbart (approved):

PR #861, opened 2026-08-10
DOCS-6394 Replace licence includes with literal notices (batch 1/6)

mariadb-stefan-hinz:
Implements [Daniel's decision on DOCS-6394](https://mariadbcorp.atlassian.net/browse/DOCS-6394): licence notices go **literally** into the `.md` source, with no include dependency, so the licence a page claims is readable from the file itself.

This is **batch 1 of 6**. Batches are split by space, as Daniel suggested.

## What Changed

242 pages across five spaces. Every diff is one line — the reusable-content include is replaced by the notice text it already resolved to:

```diff
-{% include "https://app.gitbook.com/s/.../~/reusable/pNHZQXPP5OEz2TgvhFva/" %}
+<sub>_This page is: Copyright © 2026 MariaDB. All rights reserved._</sub>
```

| Space | Pages |
|---|--:|
| analytics | 133 |
| tools | 95 |
| connectors | 7 |
| galera-cluster | 6 |
| maxscale | 1 |
| **Total** | **242** |

238 resolved to the MariaDB copyright notice, 3 to CC BY-SA / Gnu FDL. One galera-cluster page (`galera-management/general-operations/managing-sequences-in-galera-cluster.md`) carried the notice text without its `<sub>` wrapper and is normalized to the canonical form.

## No Page Changes Its Licence

The conversion is mechanical and licence-preserving. All 9,882 content pages were reclassified before and after; the notice each page claims is identical, and the conversion is idempotent (re-running it produces no diff). `242 files changed, 242 insertions(+), 242 deletions(-)` — one line per page, no whitespace churn.

## CI: the 22 Link Errors Are Pre-existing

The link-check gate scans changed files, so touching 242 pages surfaces existing rot. **This diff adds no links at all** — it only removes them.

Verified rather than asserted, using the CI-exact lychee invocation over the same 242 files in a clean worktree at unmodified `origin/main`:

|  | Links checked | Errors | Unique failing targets |
|---|--:|--:|--:|
| this branch | 2,128 | 22 | 20 |
| clean worktree at `origin/main` | 2,369 | 22 | 20 |

The failing-target lists are **identical**. The 241-link difference in the totals is exactly the `app.gitbook.com` include URLs this PR removes — so it slightly *reduces* what CI has to check.

The 20 targets are 5 `broken-reference` migration placeholders and 15 pieces of external rot (retired `shop.oreilly.com` product pages, `db.rstudio.com`, `mran.microsoft.com`, and similar).

codespell passes clean.

[...]

PR #863, opened 2026-08-10
DOCS-6394 Replace licence includes with literal notices (batch 3/6)

mariadb-stefan-hinz:
Implements [Daniel's decision on DOCS-6394](https://mariadbcorp.atlassian.net/browse/DOCS-6394): licence notices go **literally** into the `.md` source, with no include dependency, so the licence a page claims is readable from the file itself.

This is **batch 3 of 6** — the whole `server` space. See #861 (batch 1) and #862 (batch 2).

## What Changed

1,537 pages. This is the one batch that does more than swap a line, because `server` is where the *path*-include form and the marker comments live.

**1. 1,482 path includes become the literal notice** they already resolved to:

```diff
-{% include "../../../.gitbook/includes/license-cc-by-sa-gnu-fdl.md" %}
+<sub>_This page is licensed: CC BY-SA / Gnu FDL_</sub>
```

1,394 CC BY-SA / Gnu FDL, 84 copyright, 3 GPLv2-help, 1 GPLv2.

**2. 1,195 marker comments are dropped.** Nearly all the error-code pages carried both an include and a `<!-- This page is licensed: CC BY-SA / Gnu FDL -->` comment. That comment existed *because* the include form is opaque — it was how a reader or a grep could tell what the page claimed. The literal notice now says it directly, so the comment is redundant:

```diff
-{% include "../../../.gitbook/includes/license-cc-by-sa-gnu-fdl.md" %}
-
-<!-- This page is licensed: CC BY-SA / Gnu FDL -->
+<sub>_This page is licensed: CC BY-SA / Gnu FDL_</sub>
```

**3. 55 pre-existing literal variants are normalized.** Flagging this explicitly since it's adjacent cleanup rather than pure conversion: 53 pages wrote the GPLv2-help notice with the `fill_help_tables.sql` link *inside* the italics rather than beside it, and 2 pages had the notice text with no `<sub>` wrapper. Both now match the canonical form the other 491 pages use. Happy to drop these 55 from the PR if you'd rather keep it strictly mechanical.

`1,537 files changed, 1,537 insertions(+), 3,927 deletions(-)` — the extra deletions are the 1,195 marker lines plus the 1,195 blank lines that padded them.

## No Page Changes Its Licence

All 9,882 content pages were reclassified before and after; the notice each page claims is identical, and the conversion is idempotent. Unrelated includes on the same pages (e.g. `contributing-content.md`) are left untouched.

## CI

|  | Links checked | Errors | Unique failing targets |
|---|--:|--:|--:|
| this branch | 3,660 | 2 | 1 |
| clean worktree at `origin/main` | 3,657 | 2 | 1 |

The single failing target — `server/architecture/server-constraints/broken-reference`, a migration placeholder — is identical on both sides, so it is pre-existing.

**This PR does add 3 links,** and they're accounted for: three pages converted to the GPLv2-help notice, which carries a `fill_help_tables.sql` link. All three resolve OK, which is why the error count is unchanged. codespell passes clean.

### DIFF
--- a/server/architecture/server-constraints/auto_increment-constraints.md
+++ b/server/architecture/server-constraints/auto_increment-constraints.md
@@ -394,6 +394,6 @@ The offset and increment values can be configured by setting the [auto\_incremen
 
 When Galera Cluster is used, the offset and increment values are managed automatically by default. They can be managed manually by disabling the [wsrep\_auto\_increment\_control](https://app.gitbook.com/s/3VYeeVGUV4AMqrA3zwy7/reference/galera-cluster-system-variables#wsrep_auto_increment_control) system variable.
 
-{% include "../../.gitbook/includes/license-copyright-mariadb.md" %}
+<sub>_This page is: Copyright © 2026 MariaDB. All rights reserved._</sub>
 
 {% @marketo/form formId="4316" %}
--- a/server/architecture/server-constraints/foreign-key-constraints.md
+++ b/server/architecture/server-constraints/foreign-key-constraints.md
@@ -411,6 +411,6 @@ A foreign key constraint requires an index on the column. If a foreign key const
 
 When the [foreign\_key\_checks](../../ha-and-performance/optimization-and-tuning/system-variables/server-system-variables.md#foreign_key_checks) system variable is disabled, it is possible to drop the index used by a foreign key constraint. When the [foreign\_key\_checks](../../ha-and-performance/optimization-and-tuning/system-variables/server-system-variables.md#foreign_key_checks) system variable is re-enabled, InnoDB will have no way to enforce the foreign key constraint, so all operations that could potentially violate the foreign key constraint will fail.
 
-{% include "../../.gitbook/includes/license-copyright-mariadb.md" %}
+<sub>_This page is: Copyright © 2026 MariaDB. All rights reserved._</sub>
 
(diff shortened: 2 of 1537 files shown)


mariadb-stefan-hinz:
**CI result: link-check is red, and none of it comes from this diff.**

lychee exits 2 (errors found) but the step output didn't make it into the retrievable run log, so rather than guess I settled it by extracting link targets directly.

Every target lychee finds in these 1,537 files, on this branch versus a clean worktree at unmodified `origin/main`:

|  | Unique link targets |
|---|--:|
| clean worktree at `origin/main` | 2,160 |
| this branch | 2,160 |
| **introduced by this PR** | **0** |
| removed by this PR | 0 |

So the gate is reporting pre-existing rot on these pages. Running the CI-exact invocation locally over the same files gives 2 errors / 1 unique target — `server/architecture/server-constraints/broken-reference`, a migration placeholder — identical on both sides; CI additionally hits the usual runner-only external failures.

One clarification on the PR body: I noted there that the diff "adds 3 links". More precisely, it adds three *instances* of the `fill_help_tables.sql` URL (the three pages converting to the GPLv2-help notice), but that URL was already among the targets in this file set, so the number of distinct targets is unchanged at 2,160 — as the table shows.

codespell and the GitBook server-space preview pass.

PR #867, opened 2026-08-10
DOCS-6394 Add literal licence notices to unlabelled server pages (backfill 1/4)

mariadb-stefan-hinz:
Backfills pages in the `server` space that carried **no licence footnote at all**, implementing [Daniel's three rulings on DOCS-6394](https://mariadbcorp.atlassian.net/browse/DOCS-6394).

This is **backfill 1 of 4**, the second wave of the ticket. The first wave (#861–#866) converts include-based notices to literal ones; this wave adds notices to pages that never had one.

## What Changed

**444 pages** — 415 CC BY-SA / Gnu FDL, 29 MariaDB copyright. Each diff adds one line plus its blank separator:

```diff
 * [mariadb-binlog Options](mariadb-binlog-options.md)
+
+<sub>_This page is licensed: CC BY-SA / Gnu FDL_</sub>
```

## Daniel's Three Rulings, as Implemented

**1. READMEs get a notice — but only if the page is not empty.** This one has teeth: GitBook auto-generates the sub-page link list **only** when a page body is empty, so adding a notice to an empty landing page silently removes that navigation. Verified live rather than assumed — an empty README renders its children *twice* (sidebar + auto body list), a non-empty one renders them *once*. So **467 empty pages are deliberately left without a notice**, 396 of which have children and would actually have lost navigation.

**2. Pages with no sibling precedent get the MariaDB copyright notice.**

**3. READMEs of directories full of GPLv2/`fill_help_tables` pages get CC BY-SA**, since that notice asserts a provenance those landing pages do not have.

Where a precedent does exist, the notice comes from the nearest ancestor directory in which ≥80% of already-labelled siblings agree.

## Two Corrections Found While Running This

Both are why the page counts here differ slightly from the ticket:

- **A fifth notice form exists.** Connector/Python asserts **CC BY 3.0** — `<sub>_This page is_</sub> [<sub>_covered_</sub>](license.md) …` — split across several `<sub>` spans with links, so no text search for the four known notices finds it. Its `README.md` and `license.md` say the same in prose. **The 16 remaining `connectors/mariadb-connector-python/` pages are therefore held out of this backfill**, because stamping "Copyright © 2026 MariaDB. All rights reserved." on them would contradict a licence grant made on those very pages. Flagged on the ticket for a ruling. (That subtree is already inconsistent: 2 pages carry the MariaDB copyright notice, 1 carries CC BY 3.0.)
[...]

PR #866, opened 2026-08-10
DOCS-6394 Replace licence includes with literal notices (batch 6/6)

mariadb-stefan-hinz:
Implements [Daniel's decision on DOCS-6394](https://mariadbcorp.atlassian.net/browse/DOCS-6394): licence notices go **literally** into the `.md` source, with no include dependency, so the licence a page claims is readable from the file itself.

This is **batch 6 of 6** — the rest of `release-notes` (enterprise-server, columnstore, maxscale, galera-cluster, enterprise-operator, enterprise-manager, tools, advanced-cluster and the space root). See #861–#865 for batches 1–5.

**With this batch merged, no page in the repo reaches its licence notice through an include.**

## What Changed

**345 pages.** Each diff is one line — the reusable-content include is replaced by the notice text it already resolved to.

**Plus 49 pre-existing misspellings across 35 files, in a separate commit.** The codespell gate scans changed files, so touching these pages drags their existing typos into scope; fixed rather than ignored, per PR #828's precedent. Mostly old ColumnStore and Enterprise Server release notes.

Worth flagging: **codespell's suggestion was wrong in one case.** It flags `Aache` in the ColumnStore 1.1.3 notes and proposes "Cache" or "Ache"; the actual text is "Aache Kafka Data Adapter", so the correct fix is **Apache**. The other multi-candidate hits (`tread`→`thread`, `unsing`→`using`, `fter`→`After`, `doesnt`→`doesn't`) were each resolved by reading the sentence rather than taking codespell's first suggestion.

## No Page Changes Its Licence

All 9,882 content pages were reclassified before and after; the notice each page claims is identical, and the conversion is idempotent.

## CI: This PR Introduces Zero Links

Every link target lychee finds in these 345 files, on this branch versus a clean worktree at unmodified `origin/main`:

|  | Unique link targets |
|---|--:|
| clean worktree at `origin/main` | 7,153 |
| this branch | 7,151 |
| **introduced by this PR** | **0** |
| removed by this PR | 2 — the two `app.gitbook.com` reusable-include URLs |

Any link error the gate reports is pre-existing rot on these pages. codespell passes clean.

## After This Merges

Problem 1 on the ticket — the **1,951 pages that carry no licence notice at all** — is still open, and still needs the two decisions Daniel's comment didn't cover: whether `README.md` landing pages get a notice (763 of them), and which notice applies where. I've posted a concrete per-subtree proposal on the ticket for sign-off.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #870, opened 2026-08-10
DOCS-6394 Add literal licence notices to maxscale and smaller spaces (backfill 4/4)

mariadb-stefan-hinz:
Backfills `maxscale` plus the smaller spaces — pages that carried **no licence footnote at all** — implementing [Daniel's three rulings on DOCS-6394](https://mariadbcorp.atlassian.net/browse/DOCS-6394).

This is **backfill 4 of 4**, completing the second wave. See #867–#869 for backfills 1–3.

## What Changed

**340 pages** — 262 CC BY-SA / Gnu FDL, 78 MariaDB copyright, across `maxscale` (210), `general-resources` (30), `galera-cluster` (28), `platform` (24), `analytics` (23), `tools` (23) and `home` (2).

**Plus 3 pre-existing codespell hits, in a separate commit — and two of them are not spelling fixes:**

- **`ser` is not a misspelling of `set`.** Two galera upgrade pages end `"...for more information.ser"` — a stray fragment left over from the migration. codespell wanted `set`; the correct fix is to **delete** it.
- **`variante` is correct French.** It appears in a French-language book description in `general-resources/books.md` ("la principale variante de MySQL"). That goes to `.codespellignore` rather than having correct prose rewritten.

Only the third (`adviced`→`advised`) is an ordinary typo fix.

## Daniel's Three Rulings, as Implemented

**1. READMEs get a notice — but only if the page is not empty.** This one has teeth: GitBook auto-generates the sub-page link list **only** when a page body is empty, so adding a notice to an empty landing page silently removes that navigation. Verified live rather than assumed — an empty README renders its children *twice* (sidebar + auto body list), a non-empty one renders them *once*. So **467 empty pages are deliberately left without a notice**, 396 of which have children and would actually have lost navigation.

**2. Pages with no sibling precedent get the MariaDB copyright notice.**

**3. READMEs of directories full of GPLv2/`fill_help_tables` pages get CC BY-SA**, since that notice asserts a provenance those landing pages do not have.

Where a precedent does exist, the notice comes from the nearest ancestor directory in which ≥80% of already-labelled siblings agree.

## Two Corrections Found While Running This

Both are why the page counts here differ slightly from the ticket:

- **A fifth notice form exists.** Connector/Python asserts **CC BY 3.0** — `<sub>_This page is_</sub> [<sub>_covered_</sub>](license.md) …` — split across several `<sub>` spans with links, so no text search for the four known notices finds it. Its `README.md` and `license.md` say the same in prose. **The 16 remaining `connectors/mariadb-connector-python/` pages are therefore held out of this backfill**, because stamping "Copyright © 2026 MariaDB. All rights reserved." on them would contradict a licence grant made on those very pages. Flagged on the ticket for a ruling. (That subtree is already inconsistent: 2 pages carry the MariaDB copyright notice, 1 carries CC BY 3.0.)
- **The first audit walked the filesystem, not git.** It counted two untracked local `graphify-out/GRAPH_REPORT.md` artifacts as unlabelled pages. Now driven off `git ls-files`, which also normalizes the macOS-cased `mariadb-cloud/Quickstart|Security` duplicates. Corrected totals: **9,880** tracked content pages, **1,948** unlabelled (not 9,882 / 1,951).

## Notice Placement

Matches the canonical form used by the 1,325 pages that already had a literal notice: the notice, a blank line, then the `{% @marketo/form %}` block if the page has one; otherwise at the end of the file.

## Independent of the Conversion PRs

This wave and the conversion wave (#861–#866) touch **completely disjoint file sets** — 4,835 files vs 1,466, zero overlap — so they can merge in any order without conflicting.

## CI

|  | Unique link targets |
|---|--:|
| clean worktree at `origin/main` | 1,580 |
| this branch | 1,580 |
| **introduced** | **0** |
| removed | 0 |

codespell passes after the changes above.

## With This Merged

Every page in the repo that should carry a licence notice has one literally in its source. What remains uncovered is deliberate and enumerated: **467 empty pages** (a notice would break their auto-generated sub-page navigation) and **16 `connectors/mariadb-connector-python/` pages** awaiting a ruling on CC BY 3.0.


mariadb-stefan-hinz:
**CI result: link-check is red, and none of it comes from this diff.**

The gate reports 33 failing targets across the 340 changed files. This diff adds a single `<sub>` line per page and **no links at all** — neither the CC BY-SA nor the copyright notice text contains one.

Measured rather than asserted. Every link target lychee finds in these files, on this branch versus a clean worktree at unmodified `origin/main`:

|  | Unique link targets |
|---|--:|
| clean worktree at `origin/main` | 1580 |
| this branch | 1580 |
| **introduced by this PR** | **0** |

Identical sets, so every one of the 33 failures is pre-existing rot on pages this PR happens to touch. Breakdown:

- 2 — `broken-reference` migration placeholder
- 13 — dead relative link
- 15 — external rot
- 3 — retired private-repo 404 (DOCS-6385)

codespell passes.

PR #872, opened 2026-08-11
fix typo in foreign keys section

xdavidliu:
correct spelling is integrity


mariadb-stefan-hinz:
@xdavidliu Thanks for the fix!

PR #878, opened 2026-08-12
DOCS-6256 Document package-wide TYPE declarations

mariadb-stefan-hinz:
## Why this is a draft

**MDEV-39587 is not merged to `main`.** Remote `main` is `f4a85225ea1` (2026-08-05) and contains no MDEV-39587 or MDEV-13139 commit. The patch lives only on `bb-13.0-bar-MDEV-39587`, `bb-13.1-bar-MDEV-39587`, `bb-13.1-bar-MDEV-13139`, and (in an older form) `preview-13.1-preview`.

MDEV-39587 is **Stalled**: QA signed off on 2026-07-28, 15 regressions were filed against it and closed, but **MDEV-40721 (SIGSEGV) is still Open**, and the MDEV-40430 test (cross-schema package type in a trigger) is commented out in the branch. The ticket was closed as Fixed on 2026-08-11 and reopened on 2026-08-12 ("Closed in a mistake"); the branch was force-pushed on 2026-08-12.

So the content is source-verified but the **version annotation is forward-looking**. Please don't merge until the patch is in `main`, at which point the "from MariaDB 13.1" hints need re-checking — the parent task MDEV-13139 currently carries fix version 13.2.

Note also that the DOCS ticket says "Release Series: 13.0". That is stale: Sergei Golubchik moved the fix version from 13.0 to 13.1 on 2026-06-16.

## What the feature is

`MDEV-39587` allows a `TYPE` declaration in a **package specification**. Such a type is public, and routines outside the package can declare variables of it by qualified name — `pkg1.rec0_t` (two-step) or `test.pkg1.rec0_t` (three-step). Only in `sql_mode=ORACLE`: the new grammar alternative sits inside `%ifdef ORACLE`.

Verified against `bb-13.1-bar-MDEV-39587` @ `a63ab98e319` (the newest patch version), primarily against the 16 new `mysql-test/suite/compat/oracle/*sp-package-spec-type*` test/result pairs.

## Changes

**`declare-type.md`** — a new `## Package-Wide Types` section covering:

- syntax of the declaration and the two qualified reference forms
- where a package type is accepted: local variables, package routine parameters (specification *and* body), package function `RETURN`, associative-array element type, `REF CURSOR RETURN`, plus same-specification self-references to a type declared earlier
- where it is rejected, with the exact errors
- name resolution through `@@path` / `SET PATH`, and the call-time failure after the package is dropped
- cyclic package type dependencies
- the `EXECUTE ON PACKAGE` privilege requirement, and that it does not extend to the package body
- the `mariadb-dump` ordering limitation for inter-package type dependencies

**`create-package.md`** — the Oracle-mode grammar gains the `type_declaration` element (confirmed against `sql/sql_yacc.yy` on the branch vs `main`), plus a short `## Package-Wide Type Declarations` section pointing at `DECLARE TYPE`, and a See Also entry.

## Two things the ticket got wrong

The Jira description lists only one restriction — schema-level routine parameters and function `RETURN` types. The tests show two more, both documented here:

| Context | Error |
| --- | --- |
| Inside a trigger | `Incorrect usage of TRIGGER and` `` `pkg`.`rec` `` |
| Inside an event | `Incorrect usage of EVENT and` `` `pkg`.`rec` `` |

Two apparent restrictions in the tests turned out not to be restrictions, and are deliberately *not* documented: the `ER_ILLEGAL_PARAMETER_DATA_TYPES2_FOR_OPERATION` case is an `OPEN..FOR` row-count mismatch (2-field record vs 3-column `SELECT`), and the `ER_NOT_ALLOWED_IN_THIS_CONTEXT` case is the pre-existing ban on `REF CURSOR` package-level variables (`'sys_refcursor' is not allowed in this context`).

## Checks

- `codespell --check-filenames --skip ./.git --ignore-words .codespellignore` — clean
- `lychee` on both changed files — 20 total, 14 unique, **20 OK, 0 errors**
- All five new relative link targets confirmed to exist on disk

### DIFF
--- a/server/reference/sql-statements/data-definition/create/create-package.md
+++ b/server/reference/sql-statements/data-definition/create/create-package.md
@@ -40,6 +40,7 @@ package_characteristic:
 package_specification_element:
     FUNCTION_SYM package_specification_function ;
   | PROCEDURE_SYM package_specification_procedure ;
+  | type_declaration ;
 
 
 package_specification_function:
@@ -139,6 +140,26 @@ The function parameter quantifiers for `IN`, `OUT`, `INOUT`, and `IN OUT` are su
 
 `OUT`, `INOUT` and its equivalent `IN OUT`, are only valid if called from `SET` and not `SELECT`. These quantifiers are especially useful for creating functions and procedures with more than one return value. This allows functions and procedures to be more complex and nested.
 
+## Package-Wide Type Declarations
+
+{% hint style="info" %}
+This feature is available from MariaDB 13.1, in `sql_mode=ORACLE` only.
+{% endhint %}
+
+A package specification can declare data types with `TYPE`, alongside its public routines. Such a type is public: routines outside the package can declare variables of it, using the qualified name `package_name.type_name` or `schema_name.package_name.type_name`.
+
+```sql
+SET sql_mode=ORACLE;
+DELIMITER $$
+CREATE OR REPLACE PACKAGE pkg AS
+  TYPE varchar_array IS TABLE OF VARCHAR(2000) INDEX BY INTEGER;
+END;
+$$
+DELIMITER ;
+```
+
+Using the type requires the `EXECUTE` privilege on the package. For the contexts where a package type is accepted, name resolution, and the restrictions that apply, see [DECLARE TYPE](../../programmatic-compound-statements/declare-type.md).
+
 ## Examples
 
 ```sql
@@ -157,6 +178,7 @@ DELIMITER ;
 ## See Also
 
 * [CREATE PACKAGE BODY](create-package-body.md)
+* [DECLARE TYPE](../../programmatic-compound-statements/declare-type.md)
 * [SHOW CREATE PACKAGE](../../administrative-sql-statements/show/show-create-package.md)
 * [DROP PACKAGE](../drop/drop-package.md)
 * [Oracle SQL\_MODE](https://app.gitbook.com/s/aEnK0ZXmUbJzqQrTjFyb/community-server/about/compatibility-and-differences/sql_modeoracle)
--- a/server/reference/sql-statements/data-definition/create/create-package.md
+++ b/server/reference/sql-statements/data-definition/create/create-package.md
@@ -158,7 +158,7 @@ $$
 DELIMITER ;
 ```
 
-Using the type requires the `EXECUTE` privilege on the package. For the contexts where a package type is accepted, name resolution, and the restrictions that apply, see [DECLARE TYPE](../../programmatic-compound-statements/declare-type.md).
+Using the type requires the `EXECUTE` privilege on the `PACKAGE`. No privileges on the `PACKAGE BODY` are required. For the contexts where a package type is accepted, name resolution, and the restrictions that apply, see [DECLARE TYPE](../../programmatic-compound-statements/declare-type.md).
 
 ## Examples
 
--- a/server/reference/sql-statements/programmatic-compound-statements/declare-type.md
+++ b/server/reference/sql-statements/programmatic-compound-statements/declare-type.md
@@ -419,8 +419,192 @@ BEGIN
 END;
 ```
 
+## Package-Wide Types
+
+{% hint style="info" %}
+This feature is available from MariaDB 13.1, in `sql_mode=ORACLE` only.
+{% endhint %}
+
+A `TYPE` declaration can appear in a package specification. Such a type is public: routines outside the package can declare variables of it by qualifying the type name with the name of the package that declares it. Earlier, a `TYPE` declaration was confined to the block that declared it, so a record, associative array, or `REF CURSOR` type could not be shared between routines.
+
+### Syntax
+
+Declare the type in the package specification:
+
+```sql
+CREATE [OR REPLACE] PACKAGE [db_name.]package_name {AS | IS}
+  TYPE type_name IS record_definition;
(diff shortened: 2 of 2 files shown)


server/reference/sql-statements/data-definition/create/create-package.md
```diff
@@ -139,6 +140,26 @@ The function parameter quantifiers for `IN`, `OUT`, `INOUT`, and `IN OUT` are su
...
+SET sql_mode=ORACLE;
+DELIMITER $$
+CREATE OR REPLACE PACKAGE pkg AS
+  TYPE varchar_array IS TABLE OF VARCHAR(2000) INDEX BY INTEGER;
+END;
+$$
+DELIMITER ;
+```
+
+Using the type requires the `EXECUTE` privilege on the package. For the contexts where a package type is accepted, name resolution, and the restrictions that apply, see [DECLARE TYPE](../../programmatic-compound-statements/declare-type.md).
```
abarkov: To be a little bit clearer, I suggest:

Using the type requires the `EXECUTE` privilege on the `PACKAGE`. No privileges on the `PACKAGE BODY` are required.

I.e.
- write `PACKAGE` in the upper case
- mention that privileges on the `PACKAGE BODY` are not needed

server/reference/sql-statements/programmatic-compound-statements/declare-type.md
```diff
@@ -419,8 +419,192 @@ BEGIN
...
+| Parameter of a schema-level (standalone) procedure or function | `ERROR HY000: Incorrect usage of parameter_declaration and package_name.type_name` |
+| `RETURN` type of a schema-level (standalone) function | `ERROR HY000: Incorrect usage of RETURN and package_name.type_name` |
+| Inside a trigger | `ERROR HY000: Incorrect usage of TRIGGER and` `` `pkg1`.`rec0_t` `` |
+| Inside an event | `ERROR HY000: Incorrect usage of EVENT and` `` `pkg1`.`rec0_t` `` |
+
+Parameters and `RETURN` types of schema-level routines are exposed through [INFORMATION\_SCHEMA.PARAMETERS](../../system-tables/information-schema/information-schema-tables/information-schema-parameters-table.md), which cannot yet represent a package type.
+
+Two further restrictions apply to the qualified form itself:
+
+* Data type attributes cannot be applied to a qualified type. A length, `CHARACTER SET`, or `COLLATE` clause after `pkg1.rec0_t` is a syntax error.
```
abarkov: A qualified type cannot have CHARACTER SET or COLLATE only because a qualified type cannot be CHAR/VARCHAR/TEXT. So it's not because of incomplete work.
I'm not sure we need to document this.

server/reference/sql-statements/programmatic-compound-statements/declare-type.md
```diff
@@ -419,8 +419,192 @@ BEGIN
...
+* A qualified name has at most three parts. A longer chain, such as `cat1.db1.pkg1.type1`, raises `ERROR 4161 (HY000): Unknown data type`.
```
abarkov: We don't have 4 step qualified identifiers generally. This is not a specific limitation of TYPE. Perhaps this can also be removed.

server/reference/sql-statements/programmatic-compound-statements/declare-type.md
```diff
@@ -419,8 +419,192 @@ BEGIN
...
+```
+ERROR 42000: execute command denied to user 'user2'@'localhost' for routine 'db1.pkg1'
+```
+
+The check applies to any user other than the one that created the package, including users that hold `CREATE ROUTINE` in the same schema. `EXECUTE` on a package does not imply `EXECUTE` on its [package body](../data-definition/create/create-package-body.md), so calling the package's routines needs a separate grant.
+
+### Limitation on Dumps
+
+[mariadb-dump](../../../clients-and-utilities/backup-restore-and-import-clients/mariadb-dump.md) writes packages in name order. If two packages depend on each other's types, the file can define the dependent package first, and the restore then fails with `Unknown data type`. Dump and restore the `mysql.proc` table instead when packages have inter-package type dependencies.
+
```
abarkov, line 611: If the types reside in different databases, we can also suggest dump/restore them separately:
```
mariadb-dump --routines dependency_db >dependency_db.sql
mariadb-dump --routines dependent_db >dependent_db.sql
..
mariadb < dependency_db.sql
mariadb < dependent_db.sql
```

abarkov (changes requested):
Hi Stefan,

Looks very good for me. Thank you very much.

Please consider a few minor things suggested.

server/reference/sql-statements/programmatic-compound-statements/declare-type.md
```diff
@@ -419,8 +419,192 @@ BEGIN
 END;
 ```
 
+## Package-Wide Types
+
+{% hint style="info" %}
+This feature is available from MariaDB 13.1, in `sql_mode=ORACLE` only.
+{% endhint %}
+
+A `TYPE` declaration can appear in a package specification. Such a type is public: routines outside the package can declare variables of it by qualifying the type name with the name of the package that declares it. Earlier, a `TYPE` declaration was confined to the block that declared it, so a record, associative array, or `REF CURSOR` type could not be shared between routines.
```
abarkov: I suggest to rephrase here. It's not interesting what was earlier. It's interesting what's now:

A `TYPE` declaring  a record, associative array, or `REF CURSOR` can be shared between schema (i.e. standalone) and package routines.

PR #879, opened 2026-08-12
Add AdminNeo page to Graphical and Enhanced Clients

peterpp:


mariadb-stefan-hinz:
Thanks, Peter! Great page! 😄

PR #883, opened 2026-08-13
Create routine: Remove bogus note about unsupported COLLATE

vrana:
Current MariaDB works with it just fine. I just don't know since what version.

### DIFF
--- a/server/reference/sql-statements/data-definition/create/create-function.md
+++ b/server/reference/sql-statements/data-definition/create/create-function.md
@@ -54,8 +54,6 @@ By default, a function is associated with the current database. To associate the
 
 The parameter list enclosed within parentheses must always be present. If there are no parameters, an empty parameter list of `()` should be used. Parameter names are not case-sensitive.
 
-Each parameter can be declared to use any valid data type, except that the `COLLATE` attribute cannot be used.
-
 For valid identifiers to use as function names, see [Identifier Names](../../../sql-structure/sql-language-structure/identifier-names.md).
 
 ### RETURN
--- a/server/server-usage/stored-routines/stored-procedures/create-procedure.md
+++ b/server/server-usage/stored-routines/stored-procedures/create-procedure.md
@@ -89,8 +89,6 @@ The `IGNORE_SPACE` SQL mode applies to built-in functions, not to stored routine
 
 The parameter list enclosed within parentheses must always be present. If there are no parameters, an empty parameter list of `()` should be used. Parameter names are not case sensitive.
 
-Each parameter can be declared to use any valid data type, except that the `COLLATE` attribute cannot be used.
-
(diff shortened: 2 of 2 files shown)


mariadb-stefan-hinz:
Fact-checked against the server source — you're right, the sentence is wrong, and it's wrong for every version we document. Thanks for catching it.

**`COLLATE` on a routine parameter is accepted and honored:**

- **Grammar.** A parameter's type is `field_type_all_with_typedefs` (`sql/sql_yacc.yy:19768`), and the string branches take `opt_binary`, whose `binary` rule covers all four forms — `COLLATE <name>`, `CHARACTER SET x COLLATE <name>`, `COLLATE DEFAULT`, `COLLATE BINARY` (lines 7245–7310). Nothing in `LEX::sp_param_fill_definition` or the `COLUMN_DEFINITION_ROUTINE_PARAM` paths rejects a collation; those branches only adjust lengths under `sql_mode=ORACLE`.
- **Tests.** `mysql-test/main/sp-ucs2.test` has a section headed *"Testing COLLATE clause in - IN parameter"*: line 35 (IN, with charset), line 90 (IN **and** OUT), line 117 (bare `COLLATE`, no `CHARACTER SET`). The `.result` file has no `ERROR` for any of them, and `SELECT f('a')` returns `ucs2_unicode_ci` — i.e. the *declared parameter* collation, not the connection collation. So it isn't merely parsed and discarded.

**"Since what version":** you don't need a version note. The test and its clean results are present at `mariadb-10.6.0` and every tag since, and already at `mariadb-10.0.11`; the test hunk itself dates to 2009. That covers every supported release, so there's nothing to annotate.

The page also contradicted itself twice, which is nice corroboration: `create-procedure.md` has a *"Character Sets and Collations"* section stating "Procedure parameters can be declared with any character set/collation", and an example a hundred lines below the removed sentence doing exactly what it forbade — `OUT param1 CHAR(10) CHARACTER SET 'utf8' COLLATE 'utf8_bin'`.

Also confirmed the change is complete: that sentence appears exactly twice in the whole repo, and those are the two lines you removed. All five checks are green.

Approving as-is — queued for merge on our side. Reviewing it turned up one adjacent defect in the same file — the `CREATE OR REPLACE` example wrongly shows `ERROR 1304 ... already exists` for the `OR REPLACE` invocation — which is fixed separately in #890 so as not to add unrelated changes to your PR. Tracked as DOCS-6452.

mariadb-stefan-hinz:
Merged — thanks, @vrana. Good catch on a sentence that had been quietly wrong for the whole life of the page.

For anyone finding this later: the follow-up in #890 fixed the neighbouring `CREATE OR REPLACE` example, which wrongly showed `ERROR 1304 ... already exists` for the `OR REPLACE` invocation. The `CHARACTER SET 'utf8' COLLATE 'utf8_bin'` example on that page was left in place deliberately — it is correct, and it is exactly what demonstrates that the removed sentence was wrong.

PR #895, opened 2026-08-14
DOCS-6430 Fix 657 unresolvable cross-space links to the Server space

mariadb-stefan-hinz:
Fixes [DOCS-6430](https://mariadbcorp.atlassian.net/browse/DOCS-6430). Re-baselined against `main` before starting — see the [scope comment](https://mariadbcorp.atlassian.net/browse/DOCS-6430?focusedCommentId=1131206) for the full measurement.

## What's broken, and why it matters more than "a dead link"

These are raw `https://app.gitbook.com/s/SsmexDFPv2xG2OTyO5yV/...` URLs whose target no longer exists in the Server space. They don't 404. GitBook can't resolve the path, so it **degrades the link to the space root and marks it external**:

```html
<a href="https://app.gitbook.com/s/SsmexDFPv2xG2OTyO5yV">TokuDB<svg ...arrow-up-right...>
```

A reader clicking that lands on the **authenticated GitBook editor**. Links whose target is alive render normally as `/docs/server/...`. So this is reader-facing breakage plus an internal-URL leak, and **neither CI gate can see it** — lychee skips anything containing `{`, and the raw URLs sit behind auth.

## Scope

The ticket estimated ~343 links across two causes. Measured: **657 links / 308 files / 138 distinct targets**, in six spaces.

| Cause | Links | Files | Targets | Fix |
|---|---:|---:|---:|---|
| A. DOCS-5976 deletions | 288 | 219 | 27 | unlink, keep the text |
| B. The mariadb-backup folder move | 57 | 20 | 8 | repoint |
| **C. Pre-existing, previously unknown** | **312** | **106** | **103** | repoint |

Bucket A keeps its wording per the rule set on DOCS-5976 — a 10.6 release note should still say TokuDB was removed, it just shouldn't link to documentation that no longer exists. Bucket C is mostly pre-migration paths (`reference/sql-statements-and-structure/...`, `reference/mariadb-internals/.../error-codes/...`) reorganized without their inbound links being updated.

Also fixes the **2 broken relative links** in the Server space from the ticket description (`backup-lock.md`, `upgrading-from-mariadb-11-2-to-mariadb-11-3.md`).

Result: **657 → 2**. The two remaining are deliberate — `general-resources/about/readme/about-links.md`, which documents the alias syntax and contains example URLs in code fences, and a `~/reusable/<id>` include, which is reusable content by ID rather than a page path.

## Method, in case it's needed again

- **The filesystem is the wrong oracle.** GitBook derives page URLs from `SUMMARY.md` nav, not the Git path — `server-management/variables-and-modes/server-system-variables` looks dead on disk (the file lives under `ha-and-performance/`) but is a live page. A disk-based check reported **4,079** dead links; the nav-derived check reported **657**. Live probes settled it: 12 sampled dead targets all non-200, 8 sampled live targets all resolve.
- **Match by path suffix, not prefix.** These are reorganizations, so the tail is stable and the head moved. Prefix scoring picked the wrong target on all three ambiguous cases; suffix scoring got all three right.
- **The redirect layer is a free replacement oracle.** Where no page anywhere had the slug, probing the old published URL often returned a 307 naming the successor. That's where the target for all **84** `data-at-rest-encryption-overview` links came from — `data-at-rest-encryption-tde-fundamentals`, which I would not have guessed.

Repointed links are written as `{server}/...` aliases, which `expand-gitbook-aliases.yml` expands. That Action is a plain global `sed`, so it also handles the 10 links inside HTML `<a href="...">` table cells. No file in this PR is in the Action's exclusion set.

## CI will be red, and it is entirely pre-existing

**Proof that this PR introduces nothing**: extracting every URL from all 308 files before and after gives **0 URLs added, 158 removed**. This change can only shrink the link surface — it replaces editor URLs with `{server}` aliases (which lychee excludes) or deletes link syntax, leaving text.

The 308 changed files drag their whole pre-existing link surface into CI's changed-files check — 557 errors, none of them mine:

| Class | Count | Owner |
|---|---:|---|
| `github.com/mariadb-corporation/docs-{server,release-notes}` 404s | 450 (66 distinct) | **DOCS-6385** |
| Local relative/anchor misses | 49 | pre-existing |
| Root-relative `/broken/spaces/.../pages/<id>` | 32 | pre-existing |
| External rot + 429 rate-limiting | 26 | pre-existing |

I deliberately did **not** absorb the 450 `docs-server` links: they're 66 distinct targets belonging to DOCS-6385, and fixing them still wouldn't make this PR green, since the other ~107 would remain. Happy to absorb them if preferred — say the word.

[DOCS-6430]: https://mariadbcorp.atlassian.net/browse/DOCS-6430?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


mariadb-stefan-hinz:
**CI confirms the pre-existing diagnosis.** `link-check` is red, as predicted in the description, and the numbers line up with the local run almost exactly:

| | Local (CI-exact args) | CI |
|---|---:|---:|
| Total links | 28,975 | 28,975 |
| Errors | 557 | **546** |
| `docs-server` / `docs-release-notes` 404s | 450 | **450** |
| Root-relative `/broken/spaces/.../pages/<id>` | 32 | **32** |

The 11-error gap is transient `429 Too Many Requests` from `mariadb.org`, which I tripped locally by hammering it and CI did not.

Everything else is green: **codespell, expand-links, generate**, and all GitBook space previews.

Restating the innocence proof, since 546 errors is a lot to wave at: extracting every URL from all 308 changed files before and after this change gives **0 added, 158 removed**. This PR replaces editor URLs with `{server}` aliases (which lychee excludes) or deletes link syntax and leaves the text — neither operation can introduce an external URL. The 546 are the pre-existing link surface of 308 files being pulled into CI's changed-files check, and 450 of them belong to DOCS-6385.

One housekeeping note: the alias bot pushed a **merge commit** alongside its expansion commit, which would have broken the rebase merge this repo uses. The branch has been linearized with `git rebase --no-ff origin/main`, verified tree-identical before and after (`405a46a56…`), and force-pushed. It is now two clean commits.

PR #838, opened 2026-08-04
DOCS-6391 Add What's New and release notes for Enterprise Server 12.3.3-0 beta

dbart:
Creates the `release-notes/enterprise-server/12.3/` series for the Enterprise Server 12.3.2-0 beta: What's New, the release notes, and the series landing page, plus the three GitBook includes, the `SUMMARY.md` nav block, and an `all-releases.md` section.

Pages ship `hidden: true` — the release date isn't fixed yet. Removing that frontmatter flag from the three pages publishes them.

### Ralf — four open questions, all release engineering rather than editorial

1. **Exact release date** — TODO-6136 gives only "second week of August 2026" (ready end of first week).
2. **Bundled ColumnStore version** — TODO-6136 contradicts itself: its component table says `23.10`, its platform table header says "ColumnStore 25.10". ES 11.8.8-5 shipped 25.10.5.
3. **Does Red Hat UBI ship for 12.3?** Not present in TODO-6136's platform table, though ES 11.8.8-5 lists UBI 8 as the Docker image base.
4. **Advanced Cluster platform coverage** — TODO-6136's Advanced Cluster column lists el8 and SLES 15/16, but its own notes flag OpenSSL/Boost compatibility, SLES 15 compiler workarounds, and RHEL 9 ppc64le as "never enabled". Left deferred to DOCS-6353 rather than published.

Each is marked with an HTML comment at the point of use, and all four appear in the fact-check report on the ticket with their evidence.

### Notes on the ticket's five stated differences

Four are documented as given. Two needed adjusting against source:

- **Videx** — the claim is true, but for a different reason than stated. `storage/videx/` *is* in the Enterprise tree (the tree is shared with upstream); what excludes it is `PLUGIN_VIDEX NO` in `cmake/build_configurations/enterprise.cmake` plus `videx` in `PLUGINS_NOT_SUPPORTED`. The page says "not shipped" rather than "not in".
- **Backports to Enterprise Audit** — documented. MENT-2470 (`HOST:PORT` in connection records) and MENT-2471 (TLS version in `CONNECT` events), both Closed/Fixed with `fixVersion 12.3.2-0` and both verified present in `plugin/server_audit2/server_audit.cc`.

One addition beyond the ticket: MENT-2642 suppresses the `mariadb` client's GitHub call-to-action message in Enterprise Server, which is a further ES/CS difference.

### Sources

[...]

PR #905, opened 2026-08-17
DOCS-6390 Document the tls_certificate authentication plugin

mariadb-udaydubey:
Adds a reference page for the password-less TLS certificate authentication
plugin new in MariaDB Enterprise Server 12.3, plus its `SUMMARY.md` nav entry
and an Authentication Plugins card.

Jira: https://mariadbcorp.atlassian.net/browse/DOCS-6390
Upstream: MENT-2425 (Closed/Fixed, fixVersion 12.3.2-0)

### What the page covers

- Authentication from the certificate subject alone, matched byte-for-byte
  against the account's `REQUIRE SUBJECT` clause, with no password at any point.
- Why `REQUIRE SUBJECT` is mandatory, and why `REQUIRE SSL` / `REQUIRE X509`
  are insufficient on their own.
- Subject-matching pitfalls: case and DN field order both matter, and the
  subject is matched without the issuer — so scope `--ssl-ca` or add
  `REQUIRE ISSUER`.
- Account-management behaviors that do not apply to a password-less account
  (`PASSWORD EXPIRE`, `password_lifetime`, `USING`/`AS`, `SET PASSWORD`), and
  the `OR` trap where a password alone can log in.
- The two open WolfSSL limitations (MDEV-40382, MDEV-40398), which affect the
  Windows packages and are unfixed in 12.3.
- The activation option, with both accepted spellings.

### What's included
- New page: server/reference/plugins/authentication-plugins/authentication-plugin-tls-certificate.md
- Nav entry in server/SUMMARY.md
- Content-ref card in server/reference/plugins/authentication-plugins/README.md

### Reviewer notes

- **Two commits, one of them unrelated.** `61b2bfc73` fixes a one-line bug in
  `.claude/hooks/doc-lint.sh` (`mktemp -t` needs ≥3 trailing `X`s) that made the
  linter exit before running its `{% include %}` check. Happy to split it out if
  you'd rather this PR were docs-only.
- Page placement is an open question: this currently lives in the shared reference/plugins/authentication-plugins/ tree, consistent with how other ES-only plugins are placed elsewhere in the docs (e.g. authentication-with-gssapi.md links back into this tree). An alternative would be a task-oriented page under security/authentication-with-enterprise-server/. 
- Flagging for reviewer input rather than assuming.
The literal backslash in the HELP topic name (tls\_certificate) is a pre-existing rendering pattern shared with other underscore-named plugins (e.g. caching\_sha2\_password) — not something introduced by this page, out of scope here.

### Related
- Upstream: MENT-2425 (Closed/Fixed, fixVersion 12.3.2-0)
- Related ticket: DOCS-6391 (What's New / Release Notes for ES 12.3 beta), PR #838

### DIFF
--- a/server/reference/plugins/authentication-plugins/authentication-plugin-tls-certificate.md
+++ b/server/reference/plugins/authentication-plugins/authentication-plugin-tls-certificate.md
@@ -1,0 +1,139 @@
+---
+description: >-
+  The tls_certificate authentication plugin authenticates an account from its
+  TLS client certificate alone, with no password, by requiring the certificate
+  subject to match the account's REQUIRE SUBJECT clause.
+---
+
+# Authentication Plugin - tls\_certificate
+
+{% hint style="info" %}
+This plugin is available from MariaDB Enterprise Server 12.3. It is not included in MariaDB Community Server.
+{% endhint %}
+
+The `tls_certificate` authentication plugin authenticates a user from the TLS client certificate presented during the connection handshake. No password is involved at any point: the account holds no password, none is sent, and none is checked.
+
+The account is identified by the certificate's subject Distinguished Name (DN), which must be pinned on the account with a [`REQUIRE SUBJECT`](../../sql-statements/account-management-sql-statements/create-user.md) clause. The plugin refuses to authenticate an account that was not created with `REQUIRE SUBJECT`.
+
+Because the account has no password, [password validation plugins](../password-validation-plugins/simple-password-check-plugin.md) such as `simple_password_check` do not apply to it, and an account can be created while such a plugin is active.
+
+### Description
+
+`REQUIRE SUBJECT` alone is what makes this authentication rather than merely encryption. `REQUIRE SSL` guarantees only that the transport is encrypted, and `REQUIRE X509` only that the client presented some certificate signed by a trusted CA — neither identifies *which* client connected. Pinning the subject ties the account to one specific certificate identity.
+
+`REQUIRE SUBJECT` implies `REQUIRE X509`, so there is no need to specify `REQUIRE X509` separately.
+
+The plugin itself does no certificate validation. Signature checking, expiry, and revocation (CRL/OCSP) are all handled by the server's TLS layer during the handshake, before the plugin runs. The plugin's only check is that the account carries a `REQUIRE SUBJECT` clause; enforcing that clause is likewise the server's job.
+
+### Installing
+
+The plugin is built into MariaDB Enterprise Server and enabled by default. There is no `INSTALL SONAME` step, and no separate package to install.
+
+No client-side change is needed either. The plugin accepts any standard client authentication plugin, so existing clients and connectors connect without modification — they simply need to present the certificate.
+
+### Example
+
+Create an account whose subject DN matches the client certificate, and grant it as usual:
+
+```sql
+CREATE USER 'alice'@'%'
+  IDENTIFIED VIA tls_certificate
+  REQUIRE SUBJECT '/CN=alice/O=Example Ltd/C=FI';
+```
+
+`IDENTIFIED WITH` is accepted as a synonym for `IDENTIFIED VIA`.
+
+The client connects with its key pair and no password:
+
+```bash
+mariadb --user=alice \
+        --ssl-ca=/etc/ssl/ca.pem \
+        --ssl-cert=/etc/ssl/alice-cert.pem \
+        --ssl-key=/etc/ssl/alice-key.pem
+```
+
+If the certificate subject does not match, or the account has no `REQUIRE SUBJECT` clause, the connection fails with a standard access-denied error — the plugin defines no error code of its own:
+
+```
+ERROR 1698 (28000): Access denied for user 'alice'@'localhost'
+```
+
+## Matching the Certificate Subject
+
+The subject comparison is a byte-for-byte string comparison. Two consequences follow, and both bite in practice:
+
+* **Case matters.** `/CN=alice` and `/CN=Alice` are different subjects.
+* **Field order matters.** `/CN=alice/O=Example Ltd` and `/O=Example Ltd/CN=alice` are different subjects.
+
+Copy the DN exactly as the server renders it rather than retyping it. You can read the subject of a certificate with:
+
+```bash
+openssl x509 -noout -subject -in alice-cert.pem
+```
+
+{% hint style="warning" %}
+`REQUIRE SUBJECT` matches the subject only — not the issuer. An account therefore accepts any certificate with a matching subject signed by **any** CA the server trusts. If `--ssl-ca` trusts more than one CA, add a [`REQUIRE ISSUER`](../../sql-statements/account-management-sql-statements/create-user.md) clause to pin the issuer as well.
+
+On OpenSSL builds, leaving `--ssl-ca` unset makes the server trust the operating system CA store, which widens this considerably. Set [`--ssl-ca`](../../../security/encryption/data-in-transit-encryption/ssltls-system-variables.md) explicitly to the CA that issues your client certificates.
+{% endhint %}
+
+## Account Management Behavior
+
+Because these accounts authenticate with no password, several password-related features do not apply:
+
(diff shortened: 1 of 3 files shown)


server/reference/plugins/authentication-plugins/authentication-plugin-tls-certificate.md
```diff
@@ -0,0 +1,139 @@
...
+
+{% hint style="info" %}
+This plugin is available from MariaDB Enterprise Server 12.3. It is not included in MariaDB Community Server.
+{% endhint %}
+
+The `tls_certificate` authentication plugin authenticates a user from the TLS client certificate presented during the connection handshake. No password is involved at any point: the account holds no password, none is sent, and none is checked.
+
+The account is identified by the certificate's subject Distinguished Name (DN), which must be pinned on the account with a [`REQUIRE SUBJECT`](../../sql-statements/account-management-sql-statements/create-user.md) clause. The plugin refuses to authenticate an account that was not created with `REQUIRE SUBJECT`.
+
+Because the account has no password, [password validation plugins](../password-validation-plugins/simple-password-check-plugin.md) such as `simple_password_check` do not apply to it, and an account can be created while such a plugin is active.
```
vaintroub: I think you overemphasize "no password". We have plenty of passwordless plugins - gssapi, pam, unix_socket, named_pipe. Do you write the same "password check does not apply" to any  of them?

server/reference/plugins/authentication-plugins/authentication-plugin-tls-certificate.md
```diff
@@ -0,0 +1,139 @@
...
+
+### Description
+
+`REQUIRE SUBJECT` alone is what makes this authentication rather than merely encryption. `REQUIRE SSL` guarantees only that the transport is encrypted, and `REQUIRE X509` only that the client presented some certificate signed by a trusted CA — neither identifies *which* client connected. Pinning the subject ties the account to one specific certificate identity.
+
+`REQUIRE SUBJECT` implies `REQUIRE X509`, so there is no need to specify `REQUIRE X509` separately.
+
+The plugin itself does no certificate validation. Signature checking, expiry, and revocation (CRL/OCSP) are all handled by the server's TLS layer during the handshake, before the plugin runs. The plugin's only check is that the account carries a `REQUIRE SUBJECT` clause; enforcing that clause is likewise the server's job.
```
vaintroub, line 24: I think only this paragraph "plugin does no validation, only checks that REQUIRE SUBJECT exists" is meaningful,  anything above in this section is trivial/pre-existing knowledge. that could be in "REQUIRE SUBJECT" documentation if at all.

server/reference/plugins/authentication-plugins/authentication-plugin-tls-certificate.md
```diff
@@ -0,0 +1,139 @@
...
+
+If the certificate subject does not match, or the account has no `REQUIRE SUBJECT` clause, the connection fails with a standard access-denied error — the plugin defines no error code of its own:
+
+```
+ERROR 1698 (28000): Access denied for user 'alice'@'localhost'
+```
+
+## Matching the Certificate Subject
+
+The subject comparison is a byte-for-byte string comparison. Two consequences follow, and both bite in practice:
```
vaintroub, line 61: This belongs to REQUIRE SUBJECT documentation, not here. A link to REQUIRE subject documentation can be here.

server/reference/plugins/authentication-plugins/authentication-plugin-tls-certificate.md
```diff
@@ -0,0 +1,139 @@
...
+{% endhint %}
+
+When several authentication methods are combined with `OR`, the first one that succeeds wins. An account given both `tls_certificate` and a password method can therefore log in with the password alone if `REQUIRE SUBJECT` is absent. Requiring *both* a valid certificate and a password needs multi-factor authentication (several plugins combined with `AND`), which is not yet available.
+
+## Known Limitations on WolfSSL Builds
+
+{% hint style="warning" %}
+Builds that use WolfSSL rather than OpenSSL — which includes the Windows packages — are affected by two open issues. Neither is fixed in MariaDB Enterprise Server 12.3:
+
+* [MDEV-40382](https://jira.mariadb.org/browse/MDEV-40382) — `--ssl-crl` is not enforced, so a revoked client certificate still authenticates.
```
vaintroub, line 96: Pre-existing bug with wider reach than this plugin. belongs to general SSL documentation, if at all.

server/reference/plugins/authentication-plugins/authentication-plugin-tls-certificate.md
```diff
@@ -0,0 +1,139 @@
...
+* [MDEV-40398](https://jira.mariadb.org/browse/MDEV-40398) — the certificate subject separator is not escaped, so two different subjects can satisfy one `REQUIRE SUBJECT` account.
```
vaintroub, line 97: Pre-existing bug, belongs to REQUIRE SUBJECT documentation, if at all.

server/reference/plugins/authentication-plugins/authentication-plugin-tls-certificate.md
```diff
@@ -0,0 +1,139 @@
...
+
+Both are scheduled for the Q4/2026 server maintenance sprint. MDEV-40382 may be resolved after retesting; MDEV-40398 depends on a fix in WolfSSL itself and will not be worked around in MariaDB code, so it may persist beyond 12.3.
```
vaintroub: Planning deadlines need to be in the documentation?

vaintroub (commented):
I'd have to excuse myself, and maybe I'm not a good reviewer for documentation, and professionally I never dealt with this. I think documentation can be kept *much* shorter.
Good stuff - the example, ok  stuff - standard plugin variables.
Explanation of how plugin works - it is a passwordless plugin, which  needs REQUIRE SUBJECT for the user. The plugin checks for REQUIRE SUBJECT, and will fails without it .The rest is delegation the actual check to server's REQUIRE SUBJECT handling, and it is the server, that will fail authentication, if either SSL is not used, or client certificate is missing, or the certificate subject does not match.  That's enough for explanation. This would be enough, from my point of view. Anything else I'd rather omit. Passwordless plugins do not have password complexity check - I'd say a widely known fact, this existed for years prior and applies to numerous already existing passwordless plugins -  unix_socket, named_pipe, GSSAPI, PAM. WolfSSL problems or known limitations, I would not document it at all, and in case it absolutely needs documentation - it should be put in explanation for "REQUIRE SUBJECT", rather than exclusively for this new Enterprise-only plugin. Ditto the match rules of certificate SUBJECT.

mariadb-udaydubey:
@vaintroub 
Thanks for the detailed review. Alll fair points, and I appreciate you taking the time. Will update the PR once I hear back from Ralf on the scope question. Thanks again for the thorough read.

Acted on the two clearest items (out of 7 items suggested):

- Removed the password-validation-exemption sentence (line 18). I checked `unix_socket`, `named_pipe`, and `gssapi` — none of them document this either, so cutting it here matches the existing pattern. One thing worth flagging though: pam actually documents the opposite — its docs describe password validation blocking account creation as a known problem (MDEV-12321/MDEV-10457), not an exemption. So `tls_certificate` genuinely behaves differently from pam here (it does work fine with `simple_password_check` active, per Sergei's confirmation and Roel's test pass). Didn't think this page was the place to get into that, but flagging in case it's worth documenting somewhere at some point.

- Removed the sprint-timeline detail entirely (both the Q4/2026 date and the "may be resolved after retesting / will not be worked around" framing). Agreed that's internal planning context, not something that belongs in public docs. The limitations themselves and "neither is fixed in 12.3" stay — just the planning language is gone.

On the rest (comments 1, 3, 4, 5, 6) — relocating the `WolfSSL` bullets and subject-matching section into REQUIRE SUBJECT/general SSL docs. I agree with the direction, but that content doesn't exist elsewhere yet, so it'd mean creating new content (likely a separate ticket) rather than just moving text. Given the Critical priority here, I've looped in @mariadb-RalfGebhardt to weigh in on whether that's worth doing before beta ships or as a fast follow. Will hold off on further changes until I hear back from him.

mariadb-udaydubey:
@vaintroub 
Update: Ralf's given the go-ahead to ship the current version for beta, with the remaining relocation work (WolfSSL limitations, subject-matching section) tracked as a follow-up ticket rather than blocking this one. 
Thanks again for the review

PR #913, opened 2026-08-19
DOCS-6393 Add MaxScale HA deployment topologies

mariadb-tauseefkhan:
Closes DOCS-6393.

Adds a new MaxScale tutorial page covering the deployment topologies for a two-MaxScale
deployment, from Markus Mäkelä's "HA MaxScale Deployments" customer doc (#docs-talk,
2026-07-31). Companion to the DOCS-6336 page, *Failover With Multiple MaxScales*, which
covers the coordination modes; this one covers how many servers to deploy and where.

### What's here

- **New page** — `maxscale/mariadb-maxscale-tutorials/deployment-topologies-for-multiple-maxscales.md`,
  covering four topologies: 2 databases + 2 MaxScales, 3 + 2, 2 + 2 with a co-located MariaDB
  tiebreaker, and 2 Galera nodes + 2 MaxScales with a co-located arbitrator. Six Mermaid
  diagrams redrawn from the source doc, a comparison table, a worked `maxscale.cnf` for the
  tiebreaker case, and a "Choosing a Topology" section.
- **Nav** — `maxscale/SUMMARY.md`, Tutorials, alphabetical.
- **Sibling page** — the `majority_of_all` bullet under *Choosing a Mode* now points here,
  since a co-located tiebreaker is the cost-efficient way to reach the three servers that mode
  needs. Plus a `content-ref` in *See Also*.

### Claims verified against source

Every technical assertion was re-proved against MaxScale `p_develop` @ `b383abae9`, Galera
@ `a8b6b4c38`, and Server @ `5a3d98a36` before drafting. Two of them came out differently
from the source doc:

1. **The Galera case is not mechanically identical to the tiebreaker case.**
   `cooperative_monitoring_locks` is a MariaDB Monitor setting and Galera Monitor has no
   equivalent, so `garbd` votes in Galera's primary-component quorum (`pc.weight` default 1)
   rather than in a MaxScale lock majority. The topology and the cost are identical; the
   configuration is not — `garbd` has no SQL interface, so it must not appear in the monitor's
   `servers` list. The page says this explicitly instead of repeating "identical".

2. **Keeping the tiebreaker out of read routing is a service-level `servers=` list**, not
   `servers_no_cooperative_monitoring_locks`. That setting removes the server from the majority
   count — the one thing the tiebreaker exists for — and per its own reference docs still leaves
   it a valid write target.

### Not published yet — see the review comment

DOCS-6393 is gated on a docs-team decision that deployment-topology guidance belongs in the
product docs at all, which was Markus's own open question in the thread and is still
unanswered. Drafted so that call can be made against real copy rather than in the abstract.
Two questions for @markus456 in the comment below.

### DIFF
--- a/maxscale/SUMMARY.md
+++ b/maxscale/SUMMARY.md
@@ -63,6 +63,7 @@
   * [Configuring the Galera Monitor](mariadb-maxscale-tutorials/configuring-the-galera-monitor.md)
   * [Configuring the MariaDB Monitor](mariadb-maxscale-tutorials/configuring-the-mariadb-monitor.md)
   * [Connection Routing with MariaDB MaxScale](mariadb-maxscale-tutorials/connection-routing-with-mariadb-maxscale.md)
+  * [Deployment Topologies for Multiple MaxScales](mariadb-maxscale-tutorials/deployment-topologies-for-multiple-maxscales.md)
   * [Encrypting Passwords](mariadb-maxscale-tutorials/encrypting-passwords.md)
   * [Failover With Multiple MaxScales](mariadb-maxscale-tutorials/failover-with-multiple-maxscales.md)
   * [Failure-tolerant replication and failover](mariadb-maxscale-tutorials/failure-tolerant-replication-and-failover.md)
--- a/maxscale/mariadb-maxscale-tutorials/deployment-topologies-for-multiple-maxscales.md
+++ b/maxscale/mariadb-maxscale-tutorials/deployment-topologies-for-multiple-maxscales.md
@@ -1,0 +1,369 @@
+---
+description: >-
+  Compare the deployment topologies for running two MariaDB MaxScale instances
+  over the same cluster. Covers what each one costs in hardware, what it
+  survives, and how a co-located tiebreaker or Galera arbitrator buys
+  three-node safety at two-node cost.
+---
+
+# Deployment Topologies for Multiple MaxScales
+
+Running two MaxScale instances removes MaxScale as a single point of failure, and [cooperative locking](failover-with-multiple-maxscales.md) keeps the two instances from diverging the cluster. What cooperative locking cannot do is invent servers: `cooperative_monitoring_locks=majority_of_all`, the only mode that survives a network partition, needs a majority of the configured servers to be reachable. How many servers you deploy, and where, therefore decides what the pair actually survives.
+
+This page compares four topologies for a two-MaxScale deployment, from the cheapest to the ones that tolerate a partition, and shows how co-locating a third database node on a MaxScale server gets three-node safety at close to two-node hardware cost.
+
+{% hint style="info" %}
+This page assumes you have read [Failover With Multiple MaxScales](failover-with-multiple-maxscales.md). The lock arithmetic, the master lock, and the semisynchronous-replication requirement of `majority_of_all` are explained there and only referenced here.
+{% endhint %}
+
+## Comparing the Topologies
+
+| Topology | Hosts | Survives one database node down | Survives a network partition |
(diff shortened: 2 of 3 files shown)


mariadb-tauseefkhan:
@markus456 two things I'd like your call on before this publishes, plus one correction you
should look at because it contradicts your doc.

**1. Does this belong in the product docs at all?** This is your own open question from the
thread — "I'm not sure if this kind of deployment related documentation is useful". It's still
unanswered, and it's the one thing blocking DOCS-6393. I drafted the page anyway so the
decision can be made against real copy instead of in the abstract, but I'd rather not publish
deployment-topology guidance into the MaxScale space until the docs team agrees that's the
right genre for it. If the answer is no, this can become a KB article or stay a customer doc
and I'll close the ticket.

**2. Should the tiebreaker be excluded from promotion?** Your doc says only that the
co-located MariaDB "would not be used for reads". The page also recommends
`servers_no_promotion` on it, on the reasoning that a small instance sharing a host with
MaxScale is a poor failover target. That's my addition, not yours — please confirm it matches
your intent, or say if you deliberately left it promotable.

**3. A correction worth your eyes: the Galera case is not mechanically identical.** Your doc
says the arbitrator variant's "behavior is identical to the aforementioned case". The topology
and the hardware cost are, but the mechanism isn't, and I don't think the page can say
"identical" without misleading someone configuring it:

- `cooperative_monitoring_locks` only exists in MariaDB Monitor. Galera Monitor has no
  equivalent — it picks the lowest `wsrep_local_index`, which is why two MaxScales agree
  without coordinating.
- `garbd` therefore contributes nothing to a MaxScale lock majority. It votes in Galera's
  primary-component quorum instead, and divergence is prevented a layer below MaxScale.
- Practically: `garbd` has no SQL interface, so it must not be listed in the monitor's
  `servers`, whereas the MariaDB tiebreaker must be.

If I've read the Galera side wrong, tell me and I'll rework that section.

_Reviewed with Claude._

markus456:
1. That's ultimately up to you—you are the documentation team and own the information architecture. If you feel deployment-topology guidance fits better in a KB article or a standalone customer guide rather than core product docs, I trust your call. Go ahead and close out DOCS-6393 if that's the route you take.

2. Yes, absolutely. Adding `servers_no_promotion` is spot on and completely matches the intent. A lightweight tiebreaker co-located on a proxy host shouldn't handle primary workloads. Explicitly barring it from promotion avoids severe performance hits during a failover.

3. You are completely right, and honestly, that’s a brilliant catch—you're sharp to pick up on that distinction. I originally wrote "identical" as shorthand for "logically identical" in terms of the high-level quorum goal, but mechanically, they aren't the same at all, and using that word risks confusing anyone trying to configure it. 

Your technical breakdown is dead-on:
* **MariaDB Monitor:** Handles quorum directly at the proxy layer using `cooperative_monitoring_locks` over standard SQL connections.
* **Galera Monitor:** Relies on `garbd` at the cluster primary-component layer, while MaxScale nodes deterministically evaluate `wsrep_local_index`.

Since `garbd` doesn't expose an SQL interface, listing it in the monitor's `servers` configuration would definitely break things for anyone setting this up. Fantastic attention to detail here—please go ahead and rework that section to clarify the distinction and save our users from misconfiguring their monitors.

mariadb-tauseefkhan:
@mariadb-stefan-hinz Need your opinion on pointer 1(in the above comments). I beleive this one belongs in the documentation.

mariadb-stefan-hinz:
@mariadb-tauseefkhan If the question is, should the tutorial be part of the MaxScale documentation, then my answer is, yes, it should.

PR #915, opened 2026-08-19
Expand MariaDB Monitor cooperative monitoring documentation

ekorh475:
Add separate sections for both locking modes. At the start of each section, explain the scenario that the locking mode is meant for. Add more flowchars with explanations.

### DIFF
--- a/maxscale/reference/maxscale-monitors/mariadb-monitor.md
+++ b/maxscale/reference/maxscale-monitors/mariadb-monitor.md
@@ -1015,77 +1015,13 @@ Enables additional checks when a `STOP SLAVE` command times out during a cluster
 
 ## Cooperative monitoring
 
-As of MaxScale 2.5, MariaDB-Monitor supports cooperative monitoring. This means that multiple monitors (typically in different MaxScale instances) can monitor the same backend server cluster and only one will be the primary monitor. Only the primary monitor may perform _switchover_, _failover_ or _rejoin_ operations. The primary also decides which server is the primary. Cooperative monitoring is enabled with the [cooperative\_monitoring\_locks](mariadb-monitor.md#cooperative_monitoring_locks)-setting. Even with this setting, only one monitor per server per MaxScale is allowed. This limitation can be circumvented by defining multiple copies of a server in the configuration file.
+Cooperative monitoring means that multiple monitors (typically in different MaxScale instances) can monitor the same backend server cluster and agree on which one should be the primary monitor. Only the primary monitor can perform _switchover_, _failover_, _rejoin_ and other cluster operations. The primary monitor also decides which server is the primary, i.e. target of write queries. Cooperative monitoring is enabled with the [cooperative\_monitoring\_locks](mariadb-monitor.md#cooperative_monitoring_locks)-setting. This feature is meant to be used when multiple MaxScales manage the same replication cluster. It's useful even if [auto\_failover](#auto_failover) and similar features are not in use, to ensure that all monitors agree on the primary server.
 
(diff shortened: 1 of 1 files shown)


mariadb-udaydubey:
@ekorh475 Let me review it.

mariadb-udaydubey:
Thanks for the working here — this is a solid expansion. I reviewed it on two fronts: whether it addresses the three requested changes, and whether the locking claims are technically accurate. I verified the technical claims against the MaxScale source (the `mariadbmon` monitor module).

## ✅ All three requested changes are addressed

| # | Request | Status |
|---|---|---|
| 1 | Separate sections for both locking modes | ✅ New **Majority of running** and **Majority of all** sections |
| 2 | Explain the scenario each mode is meant for, at the start of each section | ✅ Both sections open with a "…is meant for situations where…" paragraph |
| 3 | More flowcharts with explanations | ✅ 2 → 6 scenario diagrams (plus the retained decision flowchart), each with a caption |

The shared intro now explains the majority formula and both calculations once, then links down to the two dedicated use-case sections. Nice structure.

## ✅ Technical accuracy — confirmed against source

I checked the locking claims against the MaxScale source and they all hold:

- **Majority formula `n_servers / 2 + 1` (rounded down)** and the worked examples (2→2, 3→2, 4→3, 5→3) match the implementation exactly (integer division).
- **`majority_of_all`** counts all configured servers minus those in `servers_no_cooperative_monitoring_locks`, independent of Down/Maintenance status — confirmed.
- **`majority_of_running`** counts only running (lockable) servers — confirmed, including the "possible with one server left" case (1/1 = majority).
- **Split-brain resistance of `majority_of_all`** (fixed total ⇒ at most one MaxScale can reach majority) and the **split-brain risk of `majority_of_running`** during a partition are both accurate.
- The **`cooperative_monitoring_locks`** (enum, default `none`) and **`servers_no_cooperative_monitoring_locks`** setting descriptions match the source, as do the lock names `maxscale_mariadbmonitor` and `maxscale_mariadbmonitor_master`.

No corrections needed on the substance.

## 🔴 Must-fix before merge

**1. Wrong datacenter named in a diagram caption.** In the "one datacenter down / three-datacenter" diagram (the last `majority_of_all` diagram), the caption reads *"Datacenter B is down…"* but the diagram shows **Datacenter C** as down (MaxScale C down, Server 3 down); Datacenter B is the secondary that stays up. This looks like a copy-paste from the previous caption (where "Datacenter B is down" is correct) — two consecutive captions currently both say "Datacenter B is down."
> Suggested: `_Datacenter C is down. It only contained one out of three servers, so the servers in the remaining datacenters can still form a majority._`

**2. Non-standard Mermaid arrow syntax.** In the `majority_of_running` split-brain diagram, the cross-datacenter edges use `.->`:
```
MXA .-> SB1
MXA .-> SB2
MXB .-> SA1
MXB .-> SA2
```
`.->` isn't valid Mermaid link syntax and may fail to render. Every other diagram in the section uses the dotted-link form `-.->`. Please change these four to `-.->` for correctness and consistency (ideally with an `|unreachable|` label to match the other partition diagrams).

## 🟡 Should-fix

**3. Content dropped from the intro.** The previous first paragraph ended with a useful operational note that's now gone:
> *"…only one monitor per server per MaxScale is allowed. This limitation can be circumvented by defining multiple copies of a server in the configuration file."*

If still accurate, worth keeping — it's a real constraint, not redundant with anything added.

**4. Accessibility: new diagrams dropped `accDescr`.** The two removed diagrams each had an `accDescr { … }` block, and the retained decision flowchart still has one. None of the 6 new diagrams do. The italic captions partly compensate, but for consistency each new `flowchart` should carry an `accDescr` (the caption text can seed it).

## Nits

- Typo: *"One the split happens"* → *"Once the split happens."*
- Grammar: *"a cluster of two servers **need** two locks"* → *"needs."*
- *"REST-API"* → *"REST API"* (the unhyphenated form is the dominant usage across the MaxScale docs; the diff introduced the hyphen).
- The first `majority_of_running` caption is hard-wrapped across four physical lines inside the `_…_`; every other caption is a single line — suggest unwrapping.
- `auto_failover` appears bare in that same caption, whereas prose elsewhere uses the linked `[auto\_failover]` form.

Overall: the content is accurate and the restructure fully answers the review. Just the caption error (#1) and the Mermaid arrows (#2) are blocking, since both are visible-output defects.

PR #894, opened 2026-08-14
DOCS-5976 Retire eight obsolete pages, rewrite merge-tree howto (Tier C)

mariadb-stefan-hinz:
Tier C of [DOCS-5976](https://mariadbcorp.atlassian.net/browse/DOCS-5976). Tiers A and B are merged and their redirects are live; this is the last tier that is docs work. (Tier D — the 634-page MaxScale version archive — is split out as DOCS-6435 and is On Hold pending Marketing buy-in and a MaxScale supported-versions statement.)

## Removed (8 pages)

| Page | Why |
|---|---|
| Merging from MySQL (obsolete) | bzr-era; self-declared obsolete |
| Merging New XtraDB Releases (obsolete) | XtraDB absent from the server tree |
| Merging TokuDB (obsolete) | TokuDB absent from the server tree |
| MariaDB 5.5.33 Debian and Ubuntu Installation Issues | a 2013 packaging bug |
| MariaDB Debian Live Images | self-declared obsolete |
| Moving from MySQL to MariaDB in Debian 9 | Debian 9 is EOL |
| SHOW INNODB STATUS (removed) | removed in MariaDB 5.5 |
| Install Cassandra on Fulltest VMs | the Cassandra engine is gone |

## Rewritten, not removed (1 page)

**Creating a New Merge Tree** carried the note *"This article is obsolete. We don't use bzr anymore. This howto needs to be rewritten to explain how to create a merge tree in git."*

It was on the removal list, but its companion page **Merging with a Merge Tree** is current and git-based, and opens by depending on it. Deleting the prerequisite would have orphaned a page that still works, so the howto is rewritten for git instead.

The git procedure is reconstructed from evidence, not inferred from the bzr text:

- `MariaDB/mergetrees` is live, with 14 branches on a `merge-<package>` convention.
- The branches are genuinely orphans — `git merge-base merge-pcre merge-zlib` reports no common ancestor, and each has its own root commit.
- Their layout mirrors the server tree (`merge-pcre` is `pcre/`, `merge-myrocks` is `storage/`), which is what the companion page's per-package instructions assume.
- The worked example uses the same `pcre-8.34` tree the old page described creating. Its root commit `8958f8b` is present in the server repository, merged as `10512b5630f "pcre-8.34 mergetree initial merge"` (Sergei Golubchik, 2014-06-05) — **a merge commit whose tree is byte-identical to its first parent**, which is what makes the null-merge step in the rewrite a description of what was actually done rather than a suggestion.

[...]

PR #917, opened 2026-08-19
DOCS-6435 Retire MaxScale 21.06 and 22.08 documentation

mariadb-stefan-hinz:
Step 2 of the plan agreed with Johan Wikman on [DOCS-6435](https://mariadbcorp.atlassian.net/browse/DOCS-6435?focusedCommentId=1131921): retire the two MaxScale series that are past end of life, leaving a stub page for the links that still point at them.

**224 pages removed, 37 inbound links retargeted, 1 page added.**

## Why these two series

Daniel Bartholomew confirmed from release management that 23.02, 23.08, 24.02, 25.01 and 25.10 are all still receiving maintenance updates, and that **21.06 and 22.08 are the only two that are not**. Those two are what this PR removes; the four surviving archived versions are untouched and move to `maxscale-old-versions` in a later step.

## The stub page

New: `maxscale/documentation-no-longer-available.md` — *Documentation No Longer Available*. It states that 21.06 and 22.08 are past end of life, points at the release notes that are kept as a historical record, and links the current documentation and the archive of still-supported versions.

It is marked `hidden: true`, so it stays reachable by URL — and therefore valid as a redirect destination — but does not appear in the sidebar or in search. Nobody should ever browse *to* a page announcing that documentation is gone. It sits at the space top level on purpose, so it survives the `maxscale-old-versions` restructure in step 3.

Per Johan's instruction on the ticket, links from release notes are **not** repointed at current documentation — *"that would be to rewrite the history; links adjusted to point to the current documentation would simply not tell the truth"* — so they go here instead.

## How the 37 inbound links were handled

The ticket estimated 4 + 3 links into these two trees. That figure counted only editor-form URLs and missed the release notes; the real number is **37 across 34 files**, plus 224 nav lines in `maxscale/SUMMARY.md`.

| Where the link lives | Count | Treatment |
|---|---|---|
| Release notes (21.06.x, 22.08.x, changelogs) | 26 | → the stub. History, must not be repointed. |
| Live pages with a current equivalent | 2 | → the current page: Galera Monitor, and MariaDB Monitor `#cooperative-monitoring`. |
| Live pages with no current equivalent | 2 | Unlinked, text kept: the csmon note in the 25.01 upgrade guide, and the `maxadmin`-era CDC tutorial reference in ColumnStore streaming adapters. |
| Sibling archive pages of surviving versions | 7 | → the equivalent page **in their own version**, not the stub. |

That last row is worth a note: the four `disable_master_failback` links and three `[CDC system]` links in the 23.02–25.01 configuration and CDC-connector pages all pointed at the *21.06 folder root*, which was a migration artifact rather than a real target. Repointing them inside their own version is both cheaper and more correct than routing them through the stub. Those seven also picked up a drive-by fix: `[MariaDB MaxScale](…)[CDC system](…)` had the two links jammed together with no space.

`grep` for either retired tree is now **0 repo-wide**.

## Redirects

A CSV of all **224** retired URLs → the stub is on the ticket. The URL derivation was checked against the live site (four sample paths, including the literal dot in `22.08`, all return 200), and all five destinations the stub links to return 200.

**The redirects must be applied after this merges and the GitBook sync completes**, not before — applying them first would shadow pages that are still live.

## CI

**Green.** codespell, `expand-links`, all ten GitBook checks and `link-check` pass; `SUMMARY.md` has 0 dangling nav entries after the 224 lines came out.

Getting `link-check` there took a drive-by fix. The archived 25.01 configuration guide cited `https://gnutls.org/manual/html_node/Priority-Strings.html`, which had been failing since 2026-08-14 and had already blocked PR #894 and PR #896. That is not a transient outage: gnutls.org still resolves in DNS but refuses connections on both 80 and 443, and `www.gnu.org/software/gnutls/` 302-redirects straight to it, so there is no live rendered GnuTLS manual left anywhere. The project develops at [gitlab.com/gnutls/gnutls](https://gitlab.com/gnutls/gnutls).

All 11 gnutls.org links in the repo are repointed, across 5 files:

| Old | New | Count |
|---|---|---|
| `https://www.gnutls.org/` (label *GnuTLS*) | `https://gitlab.com/gnutls/gnutls` | 9 |
| `…/manual/html_node/Priority-Strings.html` | the *Priority Strings* node of the manual source, pinned to tag `3.8.13` so the line anchor stays valid | 2 |

Two of the five files are outside the MaxScale space (`server/security/encryption/`, `server/server-management/`), so this unblocks those pages for any other PR that touches them. The trade-off on the second row is that a `.texi` source file is a worse read than the rendered manual was; a Wayback snapshot would render better but is a worse citation. Say if you would rather have the snapshot.

## Not done here, deliberately

Asset cleanup. `maxscale/.gitbook/assets/` is a flat shared pool with no per-version directories, so working out which images are now orphaned needs a full reference scan across the surviving versions, and deleting a shared asset would break live pages. Worth its own pass if we want it.

[DOCS-6435]: https://mariadbcorp.atlassian.net/browse/DOCS-6435?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


mariadb-stefan-hinz:
CI settled. **codespell: pass. expand-links: pass. All 10 GitBook checks: pass. link-check: red on the pre-existing gnutls.org failure only.**

One correction to the PR body: CI classifies it as a **timeout**, not an error — the summary reads `🚫 Errors 0` / `⏳ Timeouts 1`, which is the signature of an unreachable host rather than a broken link. Same file, same URL at line 1658, unchanged on `origin/main`.

Note the three checks came back `action_required` after the alias bot pushed its commit and had to be approved manually before they would run.

PR #920, opened 2026-08-20
DOCS-6478 Restore the swallowed connectionTimeZone value

mariadb-stefan-hinz:
Fixes [DOCS-6478](https://mariadbcorp.atlassian.net/browse/DOCS-6478). Reported by a reader.

## The defect

The `connectionTimeZone` bullet in the Connector/J timezone section lists only two of the option's three values:

> * connectionTimeZone: (LOCAL | SERVER | ) - This option defines the connection's time zone. LOCAL retrieves the JVM's default time zone, SERVER fetches the server's global time zone upon connection creation, **and allows specifying** a server time zone without requesting it during connection establishment.

The third value was written as a literal `<user-defined time zone>`. Markdown parses that as an HTML tag and drops it, so it vanished **twice** in the same bullet — once after the last pipe, and once mid-sentence, which is where the ungrammatical "and allows specifying" comes from.

## The fix

Wrap the value in a code span so it renders, in both places:

```
* connectionTimeZone: (LOCAL | SERVER | `<user-defined time zone>`) - This option defines ... and `<user-defined time zone>` allows specifying ...
```

## Verified against source

`mariadb-connector-j/src/main/resources/driver.properties:85` carries the intended text verbatim, `<user-defined time zone>` included. Corroborated by:

* `Configuration.java`, `Builder.connectionTimeZone` javadoc — "possible value are `LOCAL|SERVER|user-defined time zone`".
* `client/impl/StandardClient.java`, `handleTimezone()` — null or `LOCAL` uses `TimeZone.getDefault()`; `SERVER` queries `@@time_zone` / `@@system_time_zone`; **anything else is taken verbatim as a zone id** and parsed as a `ZoneId`, fixed-offset forms such as `+02:00` included.

The page's own pre-3.4 compatibility example already uses such a value (`connectionTimeZone=America/Los_Angeles`), so the restored text is consistent with the rest of the section.

## Files

* `connectors/mariadb-connector-j/about-mariadb-connector-j.md` — the page the reader reported
* `release-notes/connectors/java/3.4/3.4.0.md` — the same bullet, copied into the CONJ-1171 section of the 3.4.0 release note

## Pre-PR checks

* codespell (CI-exact invocation) — clean
* lychee (CI-exact excludes, both files) — 84 OK, 0 errors

[DOCS-6478]: https://mariadbcorp.atlassian.net/browse/DOCS-6478?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/connectors/mariadb-connector-j/about-mariadb-connector-j.md
+++ b/connectors/mariadb-connector-j/about-mariadb-connector-j.md
@@ -866,7 +866,7 @@ The simplest approach to avoid time zone headaches is for the client and server 
 
 There are 3 options that control timestamps behavior in the java connector:
 
-* connectionTimeZone: (LOCAL | SERVER | ) - This option defines the connection's time zone. LOCAL retrieves the JVM's default time zone, SERVER fetches the server's global time zone upon connection creation, and allows specifying a server time zone without requesting it during connection establishment.
+* connectionTimeZone: (LOCAL | SERVER | `<user-defined time zone>`) - This option defines the connection's time zone. LOCAL retrieves the JVM's default time zone, SERVER fetches the server's global time zone upon connection creation, and `<user-defined time zone>` allows specifying a server time zone without requesting it during connection establishment.
 * forceConnectionTimeZoneToSession: (true | false) - This setting dictates whether the connector enforces the connection time zone for the session.
 * preserveInstants: (true | false) - This option controls whether the connector converts Timestamp values to the connection's time zone.
 
--- a/connectors/mariadb-connector-j/about-mariadb-connector-j.md
+++ b/connectors/mariadb-connector-j/about-mariadb-connector-j.md
@@ -866,7 +866,7 @@ The simplest approach to avoid time zone headaches is for the client and server 
 
 There are 3 options that control timestamps behavior in the java connector:
 
-* connectionTimeZone: (LOCAL | SERVER | `<user-defined time zone>`) - This option defines the connection's time zone. LOCAL retrieves the JVM's default time zone, SERVER fetches the server's global time zone upon connection creation, and `<user-defined time zone>` allows specifying a server time zone without requesting it during connection establishment.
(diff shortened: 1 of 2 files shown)


mariadb-stefan-hinz:
## Follow-up commit: document the value format

Restoring the placeholder said *that* there is a third value but not what may go in it, so the bullet now names the format, gives both accepted shapes, and states the failure mode:

> A user-defined value must be a valid [Java ZoneId](https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/time/ZoneId.html): either a region ID such as `Europe/Berlin` or `America/Los_Angeles`, or a fixed offset such as `+02:00`. An unrecognized value fails the connection with an `Unknown zoneId` error.

**Verified against `StandardClient.handleTimezone()`**, which parses any value other than `LOCAL` / `SERVER` as:

```java
try {
  context.setConnectionTimeZone(TimeZone.getTimeZone(ZoneId.of(zoneId).normalized()));
} catch (DateTimeException e) {
  try {
    context.setConnectionTimeZone(
        TimeZone.getTimeZone(ZoneId.of(zoneId, ZoneId.SHORT_IDS).normalized()));
  } catch (DateTimeException e2) {
    throw new SQLException(String.format("Unknown zoneId %s", zoneId), e);
  }
}
```

So region IDs and fixed offsets both work (the `SHORT_IDS` fallback also accepts the deprecated three-letter ids such as `PST`, which is deliberately not advertised), and the error text is quoted from the source rather than paraphrased.

**Form of the entry** mirrors `passwordCharacterEncoding` on the same page, which already links the relevant Java class and gives an example — so this introduces no new convention. The `docs.oracle.com/en/java/javase/11/...` link form matches the three other Java API links in the 3.x prose on this page; `ZoneId` is unchanged since Java 8, which is the driver's baseline (`<source>1.8</source>` in `pom.xml`). The URL was checked by hand (200) because `docs.oracle.com` is on the lychee exclude list, so CI does not verify it.

**Not applied to `release-notes/connectors/java/3.4/3.4.0.md`** — that page records what the 3.4.0 release said and is not a reference for the option, so it keeps the rendering fix only.

PR #926, opened 2026-08-20
DOCS-6485 Correct the MDEV-32932 release-note bullet

mariadb-stefan-hinz:
Fixes the MDEV-32932 bullet in the five release notes that carry it. Daniel Bartholomew approved correcting the released notes in place ([ticket comment](https://mariadbcorp.atlassian.net/browse/DOCS-6485)).

## The defect

The bullet said MDEV-32932 *"adds support for `BACKUP STAGE` to `mariadb-backup`"*. It does not — `mariadb-backup` has issued `BACKUP STAGE` statements since **10.4.1**, five and a half years earlier. What 10.11.8 actually changed is where the work happens, and the practical benefit (a much shorter commit-blocking window) was missing from the note entirely.

Verified at the tags, `extra/mariabackup/`:

| Tag | Statements issued |
|---|---|
| `mariadb-10.4.0` | none — `FLUSH TABLES WITH READ LOCK`; `sql/backup.cc` does not exist |
| `mariadb-10.4.1` … `mariadb-10.11.7` | `START` + `BLOCK_COMMIT` + `END` |
| `mariadb-10.11.8` | `START` + `FLUSH` + `BLOCK_DDL` + `BLOCK_COMMIT` + `END` |

At 10.11.7 `START` and `BLOCK_COMMIT` are issued back to back in one function (`backup_mysql.cc:885-887`), so essentially all copying happens with commits blocked. At 10.11.8 each stage is its own function called at the right point (`xtrabackup.cc:4912-5130`): `stage_start` copies InnoDB and online Aria data, `stage_flush` the not-in-use tables of other engines, `stage_block_ddl` the rest plus `ddl_log::backup()`, leaving only log-table and stats tails under `BLOCK_COMMIT`.

`--rsync` and `--no-backup-locks` at 10.11.8 are bound to `ignored_option`, documented as "Obsolete depricated option", and print `Obsolete option: <name>. Ignored` — so "obsoletes" was accurate and is kept.

## Also normalized

The five copies of one bullet used **two product names and three link forms** for the same targets — `mariadb-backup` vs `MariaDB Backup`, and `app.gitbook.com/s/…` vs `app.gitbook.com/o/…/s/…` vs public `mariadb.com/docs/server/…`. All five bullets are now byte-identical, with all four links on the `{server}` alias.

## Drive-by fixes

- `mariadb-backup-and-backup-stage-commands.md:16` — dropped "as do all releases in the 10.6 series and earlier", a claim I introduced in DOCS-6484. It over-reaches: 10.4.0 and earlier have no `BACKUP STAGE` at all, so they don't use the older behavior that section describes, they use FTWRL. The preceding clause already covers everything before the listed releases.

## Not changed, flagged instead

The sibling Backup bullets in 11.0.6, 11.1.5 and 11.2.4 (three per file) still say "MariaDB Backup" and link to public `mariadb.com/docs/server/…` URLs, where 10.11.8 says `mariadb-backup` with GitBook links. Those links work, so this is a naming/style inconsistency rather than a defect, and it was outside the ticket — worth a separate decision by the release-notes owner.

Changelog pages are out of scope: they are generated from commit titles.

## Checks

`doc-lint.sh` (codespell + lychee, CI-mirroring flags) exits 0 on all six changed files, with both tools confirmed installed locally.


dbart:
Looks good to me. Approved.

PR #944, opened 2026-08-22
Add LibreDB Studio to Graphical and Enhanced Clients

kaya-abdullah:
Adds a page for LibreDB Studio to the Graphical and Enhanced Clients section, with
the matching content-ref in the section README and the nav entry in
server/SUMMARY.md.

LibreDB Studio is an open source (MIT) SQL IDE used through a web browser, deployed
next to the database as a container, a Helm chart, an OpenShift operator or an npm
package. It connects to MariaDB over the MySQL protocol with the mysql2 driver. I am
the author.

The page follows adminneo.md as its template, including the third-party notice and
the license footer. Two things I checked against a live server before writing it:

* Schema browsing, query execution, ER diagrams, schema comparison and
  EXPLAIN FORMAT=JSON plans work against MariaDB 12.3.
* The metrics and slow query views read performance_schema, which MariaDB ships
  disabled by default. The page says so rather than leaving a reader with empty
  panels.

The screenshot is the entity-relationship diagram, generated from a MariaDB 12.3
schema. Happy to drop the Marketo form block if that is added on your side rather
than by contributors.

This contribution is donated under the MariaDB Contributor Agreement.


mariadb-stefan-hinz:
Perfect page – thanks for adding, Abdullah!

PR #961, opened 2026-08-25
DOCS-6519 Correct the READ COMMITTED isolation level section

mariadb-stefan-hinz:
Fixes [DOCS-6519](https://mariadbcorp.atlassian.net/browse/DOCS-6519).

Michael "Monty" Widenius flagged three claims about `READ COMMITTED` as false in `#server-dev` on 2026-08-24. Two of the three turned out to be in our docs near-verbatim, and checking them against the source surfaced four more defects on the same page.

Every claim below was verified against `mariadb-server` at `main` (13.1.0-alpha); the binary-logging example was executed on a live 12.3.3 server.

## The page contradicted itself about gap locking

Line 79 said that at `READ COMMITTED`, range searches lock the index range scanned "using gap locks or next-key (gap plus index-record) locks". The hint three lines below said "there is no InnoDB gap locking". The hint was right:

```
storage/innobase/row/row0sel.cc:4776

  /* Do not lock gaps at READ UNCOMMITTED or READ COMMITTED
  isolation level */
  const bool set_also_gap_locks =
    prebuilt->select_lock_type != LOCK_NONE
    && trx->isolation_level > TRX_ISO_READ_COMMITTED
```

Line 79 was inherited MySQL manual prose describing the pre-10.5 `innodb_locks_unsafe_for_binlog=OFF` world. It is gone; the section now makes one statement, under a `Gap Locking at READ COMMITTED` heading.

## Foreign key checks were the wrong exception

The old text named "foreign-key constraint checking and duplicate-key checking" as the two exceptions. Foreign key checks skip gap locks at `READ COMMITTED` in MariaDB — `row_ins_check_foreign_constraint()` sets `skip_gap_lock = (trx->isolation_level <= TRX_ISO_READ_COMMITTED)` (`row0ins.cc:1479`), then skips the supremum lock, downgrades `LOCK_ORDINARY` to `LOCK_REC_NOT_GAP`, and drops the `LOCK_GAP`.

Duplicate-key checking is the only real exception: `row_ins_scan_sec_index_for_duplicate()` hardcodes `const ulint lock_type = LOCK_ORDINARY;` (`row0ins.cc:2122`) at every isolation level.

## The semi-consistent read hint has been wrong by default since 11.6.2

Semi-consistent read is disabled at `READ COMMITTED` whenever `innodb_snapshot_isolation` is `ON`:

```
storage/innobase/row/row0sel.cc:5405

      /* ... As read view wasn't opened, do locking read
      instead of semi-consistent one for READ COMMITTED. */
      || (trx->snapshot_isolation && trx->isolation_level
        == TRX_ISO_READ_COMMITTED )) {
```

That arrived in 10.6.19 (MDEV-34108), and `innodb_snapshot_isolation` has defaulted to `ON` since 11.6.2 (MDEV-35124). So on default settings the old hint described `READ UNCOMMITTED` only. The narrower surviving claim — a record lock released when the row turns out not to match the `WHERE` condition, unless this transaction modified it (`unlock_row()` / `row_unlock_for_mysql()`) — is now stated separately and scoped correctly.

## Row-based binary logging is not required

The old text said `READ COMMITTED` requires row logging. `binlog_format` defaults to `MIXED` (`sql/sys_vars.cc:718`) and the InnoDB check fires only on `BINLOG_FORMAT_STMT` (`ha_innodb.cc:16654`). Reproduced on 12.3.3 — `STATEMENT` fails with error 1665, `MIXED` and `ROW` both succeed — and the failing case is now shown as a worked example.

## innodb_locks_unsafe_for_binlog was presented as a live variable

Removed in 10.5.0 (MDEV-19544). Our own reference page already records "Removed: MariaDB 10.5.0". Both mentions on `set-transaction.md` are gone, as are the two on `innodb-lock-modes.md`.

## innodb_snapshot_isolation was documented twice, with opposite defaults

Line 118 said "enabled by default"; the *Current* tab of the section below said "defaults to `OFF` for backwards compatibility". The latter was stale and is gone, along with a now-empty version tab. The heading also read `innodb_snapshop_isolation` — no page linked to the misspelt anchor, so renaming it breaks nothing (confirmed with `fragcheck.py new main`).

The version boundary is now stated once: enabled by default from 11.6.2, added defaulting to `OFF` in 10.6.18, 10.11.8, 11.0.6, 11.1.5, 11.2.4, and 11.4.2.

## Drive-by fixes

`server/server-usage/storage-engines/innodb/innodb-lock-modes.md` claimed that "if a statement uses all the columns of a unique index to search for unique row, gap locking is not used". MariaDB stopped relaxing gap locking for unique indexes in 11.0 / 10.11.2 / 10.6.12 and friends — which `set-transaction.md` already documented in its *Current* REPEATABLE READ tab, so the two pages disagreed. The source comment is explicit: "Set next-key lock both for delete- and non-delete-marked records for unique search, because non-delete-marked record can be marked as deleted while transaction suspends" (`row0sel.cc:5334`). The only surviving relaxation is the narrow clustered-index `>=`-exact-match case, not any unique search.

`agent-skills/granular/statements/mariadb-set-transaction/SKILL.md` had inherited the wrong foreign-key exception and the row-logging requirement; both corrected.

## Not in this PR

The question that started the thread — whether `READ COMMITTED` is advisable for high-concurrency workloads — still has no answer in the docs, and `mariadb-transactions-and-isolation-levels-for-sql-server-users.md:164` currently recommends it on scalability grounds that Marko contradicted in the thread (a fresh read view per statement can make `READ COMMITTED` slower than `REPEATABLE READ`). That needs his sign-off rather than source reading alone, so it is filed as [DOCS-6520](https://mariadbcorp.atlassian.net/browse/DOCS-6520) rather than guessed at here. Checking the page while filing turned up three further defects that need no sign-off — the `tx_isolation = 'READ COMMITTED'` config-file recipe on that page is not a valid startup option at all and stops the server from booting — so they are on that ticket too.

## Checks

- `codespell --check-filenames --skip ./.git --ignore-words .codespellignore` on the three changed files: clean.
- `lychee` with the `link-check-pr.yml` excludes: 43 links, 0 errors.
- `fragcheck.py new main`: no new dead heading anchors, no stolen ids.

[DOCS-6519]: https://mariadbcorp.atlassian.net/browse/DOCS-6519?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

[DOCS-6520]: https://mariadbcorp.atlassian.net/browse/DOCS-6520?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/server/reference/sql-statements/transactions/set-transaction.md
+++ b/server/reference/sql-statements/transactions/set-transaction.md
@@ -76,18 +76,35 @@ The following sections describe how MariaDB supports the different transaction l
 
 A somewhat Oracle-like isolation level with respect to consistent (non-locking) reads: Each consistent read, even within the same transaction, sets and reads its own fresh snapshot. See [innodb-consistent-read.html](https://dev.mysql.com/doc/refman/en/innodb-consistent-read.html).
 
-For locking reads (`SELECT` with `FOR UPDATE` or `LOCK IN SHARE MODE`), InnoDB locks only index records, not the gaps before them, and thus allows the free insertion of new records next to locked records. For `UPDATE` and `DELETE` statements, locking depends on whether the statement uses a unique index with a unique search condition (such as `WHERE id = 100`), or a range-type search condition (such as `WHERE id > 100`). For a unique index with a unique search condition, InnoDB locks only the index record found, not the gap before it. For range-type searches, InnoDB locks the index range scanned, using gap locks or next-key (gap plus index-record) locks to block insertions by other sessions into the gaps covered by the range. This is necessary because "phantom rows" must be blocked for MariaDB replication and recovery to work.
+#### Gap Locking at READ COMMITTED
+
+InnoDB takes no [gap locks](../../../server-usage/storage-engines/innodb/innodb-lock-modes.md#gap-locks) at this isolation level. For locking reads (`SELECT` with `FOR UPDATE` or `LOCK IN SHARE MODE`), and for `UPDATE` and `DELETE` statements, InnoDB locks only the index records it examines, not the gaps before them, and thus allows the free insertion of new records next to locked records. This applies to range-type search conditions (such as `WHERE id > 100`) as well as to unique searches (such as `WHERE id = 100`).
+
+Duplicate-key checking is the exception: when InnoDB checks a unique index for a duplicate value, it takes a next-key (gap plus index-record) lock regardless of the isolation level. [Foreign key](../../../ha-and-performance/optimization-and-tuning/optimization-and-indexes/foreign-keys.md) constraint checks, by contrast, do not take gap locks at `READ COMMITTED`.
+
+Gap locks are what block phantom rows, so without them InnoDB cannot be logged safely one statement at a time. With [binary logging](../../../server-management/server-monitoring-logs/binary-log/) enabled and [binlog\_format](../../../ha-and-performance/standard-replication/replication-and-binary-log-system-variables.md#binlog_format) set to `STATEMENT`, InnoDB rejects any statement that would write rows:
+
+```sql
+SET SESSION binlog_format = STATEMENT;
+SET SESSION TRANSACTION ISOLATION LEVEL READ COMMITTED;
+UPDATE t SET v = 9 WHERE id = 1;
+ERROR 1665 (HY000): Cannot execute statement: impossible to write to binary log
+since BINLOG_FORMAT = STATEMENT and at least one table uses a storage engine
+limited to row-based logging. InnoDB is limited to row-logging when transaction
+isolation level is READ COMMITTED or READ UNCOMMITTED.
+```
+
+The default `binlog_format` of `MIXED` is unaffected, as is `ROW`.
+
+#### Semi-Consistent Reads
+
+In a semi-consistent read, an `UPDATE` or `DELETE` statement skips a row that another transaction has locked, provided the latest committed version of that row does not match the `WHERE` condition. The statement proceeds instead of waiting for the lock, which means you might see only a partially consistent read.
 
 {% hint style="info" %}
-If the `READ COMMITTED` isolation level is used or the [innodb\_locks\_unsafe\_for\_binlog](../../../server-usage/storage-engines/innodb/innodb-system-variables.md#innodb_locks_unsafe_for_binlog) system variable is enabled, there is no InnoDB gap locking except for [foreign-key](../../../ha-and-performance/optimization-and-tuning/optimization-and-indexes/foreign-keys.md) constraint checking and
-duplicate-key checking. Also, record locks for non-matching rows are released after MariaDB has evaluated the `WHERE` condition. If you use `READ COMMITTED` or enable `innodb_locks_unsafe_for_binlog`, you must use row-based binary logging.
+At `READ COMMITTED`, semi-consistent reads apply only when [innodb\_snapshot\_isolation](../../../server-usage/storage-engines/innodb/innodb-system-variables.md#innodb_snapshot_isolation) is disabled. That variable is enabled by default from MariaDB 11.6.2, and while it is enabled, `READ COMMITTED` performs an ordinary locking read and waits for the lock. Semi-consistent reads then apply to `READ UNCOMMITTED` only.
 {% endhint %}
 
-{% hint style="info" %}
-Rows that don't match are not being locked in a so called semiconsistent read. This means you might see only a partially consistent read when the transaction isolation level is `READ COMMITTED` or `READ UNCOMMITTED`.
-
-(A semiconsistent read applies to `UPDATE` and `DELETE` statements. Those statements skip locked rows, provided the version in the current read does not match the `WHERE` condition. Also, if the latest version of a record was successfully locked, but found not to match the condition, the lock is released.)
-{% endhint %}
+Separately, and regardless of `innodb_snapshot_isolation`, if InnoDB locks a record at `READ COMMITTED` or `READ UNCOMMITTED` and then finds that the record does not match the `WHERE` condition, it releases that record lock — unless the transaction has itself modified the row.
 
 ### REPEATABLE READ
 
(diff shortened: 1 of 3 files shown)


server/reference/sql-statements/transactions/set-transaction.md
```diff
@@ -76,19 +76,36 @@ The following sections describe how MariaDB supports the different transaction l
...
+The default `binlog_format` of `MIXED` is unaffected, as is `ROW`.
+
+#### Semi-Consistent Reads
 
-(A semiconsistent read applies to `UPDATE` and `DELETE` statements. Those statements skip locked rows, provided the version in the current read does not match the `WHERE` condition. Also, if the latest version of a record was successfully locked, but found not to match the condition, the lock is released.)
+In a semi-consistent read, an `UPDATE` or `DELETE` statement skips a row that another transaction has locked, provided the latest committed version of that row does not match the `WHERE` condition. The statement proceeds instead of waiting for the lock, which means you might see only a partially consistent read.
+
+{% hint style="info" %}
+At `READ COMMITTED`, semi-consistent reads apply only when [innodb\_snapshot\_isolation](../../../server-usage/storage-engines/innodb/innodb-system-variables.md#innodb_snapshot_isolation) is disabled. That variable is enabled by default from MariaDB 11.6.2, and while it is enabled, `READ COMMITTED` performs an ordinary locking read and waits for the lock. Semi-consistent reads then apply to `READ UNCOMMITTED` only.
 {% endhint %}
```
dr-m, line 105: For the record, this is not guarded by `ha_innobase::try_semi_consistent_read()` but by the `DB_LOCK_WAIT` handling in `row_search_mvcc()`.

mariadb-stefan-hinz: Thanks — confirmed, and the distinction matters more than it looks. `ha_innobase::try_semi_consistent_read()` tests only the SQL layer's request and `isolation_level <= TRX_ISO_READ_COMMITTED`, so it still arms the optimistic read at `READ COMMITTED` regardless of `innodb_snapshot_isolation`; the exclusion is in the `DB_LOCK_WAIT` branch of `row_search_mvcc()` (`row0sel.cc:5395-5408`), which declines it at contention time rather than at statement setup.

Reading that block turned up two things the page didn't say. Both are now in #969:

- **Semi-consistent reads apply to `UPDATE` only.** The engine-side contract names both statements (`handler.h:4606`, `row0mysql.h:586`), but the only caller of `try_semi_consistent_read(1)` anywhere in the server is `sql_update.cc`, lines 816 and 925 — `git log -S try_semi_consistent_read -- sql/sql_delete.cc` returns nothing, so it has never been there. On 12.3.3, with `id = 1` locked by another session: `UPDATE t SET v = 9 WHERE v = 999` returns 0 rows in 0.9 ms, while `DELETE FROM t WHERE v = 999` hits ER_LOCK_WAIT_TIMEOUT.
- **The same guard bails on `unique_search` and `index != clust_index`.** So `WHERE id = 1 AND v = 999` waits for the lock even with `innodb_snapshot_isolation=OFF`, while the range form skips the row.

The lock release for a non-matching row is unaffected by either — `sql_delete.cc:1008` does call `unlock_row()` — so the page now draws that contrast explicitly.

If the `DELETE` asymmetry is intentional rather than an oversight, we would rather say so on the page than leave it implied. Either wording is easy at this point.

dr-m: https://bugs.mysql.com/bug.php?id=3300 was only implemented for `UPDATE`, not `DELETE`. Inside InnoDB, the two operations are very similar; a `DELETE` is an `UPDATE` of a delete-mark flag in the record header.

I was wondering if it would make sense to file a bug to remove a possibly redundant call from `sql_delete.cc`.
I see that `handler::unlock_row()` had been introduced in 2001 in MariaDB/server@d61418b94fc2f377c11775039bfe5a7106541b3a which funnily refers to Gemini, not the contemporary AI one, but a storage engine that was produced by [Nusphere](https://www.theregister.com/software/2002/11/21/mysql-nusphere-settle-gpl-contract-dispute/808964). At that time, InnoDB had already been added to the repository. The "semi-consistent read" was slightly repurposing a pre-existing interface. Given that there is a non-trivial implementation of `ha_rocksdb::unlock_row()`, I think that we’d better preserve the status quo.

mariadb-stefan-hinz: Thank you — that settles it, and the page now says so outright rather than leaving the asymmetry implied:

> Semi-consistent reads are deliberately limited to `UPDATE`: the optimization was never implemented for `DELETE`, which waits for the lock.

On the `sql_delete.cc` call, one observation in case it bears on the status-quo decision: for InnoDB it is not a no-op. `ha_innobase::unlock_row()` falls through `ROW_READ_WITH_LOCKS` to `row_unlock_for_mysql()` whenever the isolation level is `READ COMMITTED` or below, so a `DELETE` that fails the `WHERE` condition really does release the record lock. On 12.3.3, session B running `DELETE FROM t WHERE w = 999` and matching nothing leaves session C free to lock the row it scanned; without the call C would wait. So removing it would be a visible behaviour change for InnoDB, not only for RocksDB.

Reading that path also caught a mistake of ours. #969 originally said the lock release "is not restricted in either of those ways" — but `row_unlock_for_mysql()` requires `new_rec_locks == 1 && prebuilt->index->is_clust()`, and during a secondary-index scan `new_rec_locks` is 2 (`row0sel.cc:5683`), or 1 for the secondary index itself. So it shares the clustered-index restriction with semi-consistent reads; only the `UPDATE`-only and unique-search restrictions are absent. Confirmed in both directions: `DELETE FROM t WHERE v = 5 AND w = 999` driven through a secondary index holds the lock to commit, while the same predicate on a clustered scan releases it. #969 is corrected and now documents that.

PR #966, opened 2026-08-25
Replace obsolete StandardOutput=syslog with journal in systemd docs

mariadb-CalebTerry:
## Summary

Resolves an open question on the [systemd configuring](https://mariadb.com/docs/server/server-management/starting-and-stopping-mariadb/systemd/configuring) page: the "Configuring MariaDB to Write the Error Log to Syslog" section told readers to set `StandardOutput=syslog` / `StandardError=syslog`, values systemd removed in 2020.

This is the follow-up to #963, which fixed the `SysLogLevel` → `SyslogLevel` capitalization on two of the same lines. That fix was correct but narrow — the directive it corrected turned out not to be the right advice at all.

## `syslog` is obsolete, and has no per-unit replacement

`syslog` and `syslog-console` were removed from `StandardOutput=` / `StandardError=` in **systemd 246** (2020). From `NEWS`, `CHANGES WITH 246`:

> `StandardError=` and `StandardOutput=` in unit files no longer support the "syslog" and "syslog-console" switches. They were long removed from the documentation, but will now result in warnings when used, and be converted to "journal" and "journal+console" automatically.

`syslog` is no longer in the accepted-values list in `systemd.exec(5)`. Every reader on RHEL 9, Ubuntu 22.04+, or Debian 11+ who followed this page got the following in the journal on **every service start**:

```
Standard output type syslog is obsolete, automatically updating to journal.
Please update your unit file, and consider removing the setting altogether.
```

The obvious question was whether `syslog` was deliberate — the old value additionally forced forwarding to a syslog daemon, which `journal` on its own does not. It was deliberate when written, but there is no unit-level replacement. `systemd.exec(5)` now says of `journal`:

> any external, additional syslog daemons receive their log data from the journal, too, hence this is the option to use when logging shall be processed with such a daemon.

Forwarding is a **journald-wide** setting: `ForwardToSyslog=` in `journald.conf`, which upstream defaults to off (`journald.conf(5)`: "By default, only forwarding to wall is enabled"). So the page now names that as the step that actually reaches a syslog daemon, and notes that on distributions where `rsyslog` reads the journal directly via `imjournal`, no forwarding is needed.

For reference, MariaDB's shipped `support-files/mariadb.service.in` sets no `Standard*` or `Syslog*` keys at all, so the `journal` default already applies and nothing in MariaDB depended on the old value.

## `SyslogLevel=err` was a second, separate defect

The same four lines recommended `SyslogLevel=err`, which reads as a severity filter. It is not one. `SyslogLevel=` sets the default priority for output lines that carry no sd-daemon `<N>` priority prefix of their own — and MariaDB does not add one. `sql/log.cc`, `print_buffer_to_file()`:

```c
  fprintf_stderr( "%d-%02d-%02d %2d:%02d:%02d %lu [%s] %.*s%.*s\n",
```

Every line is plain `2026-01-01 0:00:00 0 [Note] ...`, so `SyslogLevel=err` tagged the **entire** error log — notes and warnings included — as priority `err`. Dropped from the recommendation, with a warning hint pointing at `LogLevelMax=` for actual filtering.

## Changes

1. **Options table (lines 141–144)** — `StandardOutput=syslog` / `StandardError=syslog` → `journal`; removed the `SyslogLevel=err` row. It had mapped `mariadbd-safe`'s `syslog` option to the same obsolete values, so the table and the section contradicted each other in exactly the way #963 had just repaired elsewhere on the page.
2. **Section prose** — rewritten around `journal` plus the `journald` forwarding step.
3. **Example replaced.** The corrected unit snippet would have consisted entirely of defaults (`journal` is the default, and `SyslogFacility=` "Defaults to daemon"), so a reader could follow it, restart, and see no change. The example now shows the `journald.conf.d` drop-in that actually alters behavior, plus `SyslogIdentifier=` for the multi-instance case the following paragraph already discussed.
4. **Two hints added** — the systemd 246 removal with the exact warning text readers will have seen, and the `SyslogLevel=` clarification.

`SyslogFacility=daemon` is kept. It restates systemd's default, but it is harmless, self-documenting, and the page now explains what it selects.

## Test plan

- `grep -rn 'Standard\(Output\|Error\) *= *syslog' --include=*.md .` → no matches (was 6 lines, all in this file)
- GitBook hint blocks balanced (7 `{% hint %}` / 7 `{% endhint %}`), code fences even
- Frontmatter unchanged and present; **no headings changed**, so no anchors move — the intra-page link on line 141 still resolves
- New external link verified: `https://www.freedesktop.org/software/systemd/man/journald.conf.html` → `200`
- **codespell and lychee did not run locally** (neither installed on this machine), and `fragcheck.py` failed on a network fetch. CI is the first real run of those gates on this diff.

## Sources

- systemd `NEWS`, `CHANGES WITH 246`
- systemd `man/systemd.exec.xml` — `StandardOutput=`, `SyslogFacility=`, `SyslogLevel=` (`main`)
- systemd `man/journald.conf.xml` — `ForwardToSyslog=` (`main`)
- systemd PR #15812
- MariaDB `server` 11.4 — `sql/log.cc`, `support-files/mariadb.service.in`


mariadb-stefan-hinz (commented):
Fact-checked this against systemd's own sources on `main` (`man/systemd.exec.xml`, `man/journald.conf.xml`, `NEWS`, `src/core/load-fragment.c`) and against the MariaDB tree at 13.1.0, and ran the CI gates locally.

**The core thesis is correct and every quotation in the description is verbatim-accurate.** Two things should be fixed before merge, plus four one-line nits.

## What checks out

| Claim | Verified against |
| --- | --- |
| `syslog`/`syslog-console` removed in systemd 246 | `NEWS`, CHANGES WITH 246 — quote is verbatim |
| `syslog` no longer an accepted value | `systemd.exec.xml` `StandardOutput=` — list is `inherit, null, tty, journal, kmsg, journal+console, kmsg+console, file:, append:, truncate:, socket, fd:` |
| Exact warning text | `src/core/load-fragment.c`, `config_parse_exec_output()` — matches character-for-character. Note both `StandardOutput=` and `StandardError=` route through that same parser, so a unit with both set emits it twice |
| "any external, additional syslog daemons receive their log data from the journal" | `systemd.exec.xml`, `journal` value — verbatim |
| `SyslogFacility=` defaults to `daemon` | `systemd.exec.xml` — "Defaults to `daemon`" |
| `SyslogLevel=` is a default, not a filter | `systemd.exec.xml` — "The default syslog log level … individual lines … may be prefixed with a different log level". Defaults to `info` |
| MariaDB emits no `<N>` priority prefix | `sql/log.cc`, `print_buffer_to_file()` — plain `fprintf(stderr, …)`; level renders as `ERROR`/`Warning`/`Note` |
| `mariadb.service.in` sets no `Standard*`/`Syslog*` keys | grep across `support-files/` **and** `debian/` — zero hits |
| `ForwardToSyslog=` defaults off | `journald.conf.xml` — "By default, only forwarding to wall is enabled" |
| `/etc/systemd/journald.conf.d/*.conf` drop-in path | `journald.conf.xml` |
| `#ForwardToSyslog=` and `#LogLevelMax=` anchors resolve | `man/custom-html.xsl` derives the id from `term[1]`; both are first terms |

Gates run here (this machine has the two tools that were missing on yours): codespell **0**, lychee **0**, `doc-lint.sh` **exit 0**, `fragcheck.py check` **30 fragment links / 0 dead**. One caveat on the test plan: `www\.freedesktop\.org` is in lychee's exclude list, so CI will never check the new `journald.conf.html` link. I fetched it directly — it is live.

## 1. Please keep the `SyslogLevel=err` row in the options table

This is the one real regression. That table is an *equivalence* table — its own intro says "If an option is equivalent to a common `mariadbd-safe` option, then that is also listed." And `mariadbd-safe --syslog` does precisely this, in `scripts/mysqld_safe.sh`:

```sh
cmd="$cmd 2>&1 | logger -t '$syslog_tag_mysqld' -p daemon.error"
```

Facility `daemon`, priority `err`, applied to every line. So `SyslogFacility=daemon` + `SyslogLevel=err` was the exact systemd equivalent of `--syslog`. The row that survived is a no-op restatement of a systemd default; the row that was dropped is the one carrying information. Removing it leaves the `syslog` row mapping a non-default `mariadbd-safe` option to nothing but defaults.

There's a stronger version of your own argument available here: `mariadbd-safe --syslog` has exactly the "everything is tagged `err`, notes included" behaviour the new warning hint objects to. Saying so makes the hint land harder, and it argues for keeping the row with a Comments note rather than deleting it.

## 2. `LogLevelMax=` cannot do what the warning hint says

The hint's own premise defeats its recommendation. Because MariaDB attaches no per-line priority, every line carries `SyslogLevel` — `info` by default. `LogLevelMax=` filters on that same value, so it is all-or-nothing: `LogLevelMax=info` keeps everything, and anything more severe drops everything *including real errors*. No setting keeps errors and discards notes.

Sending a reader there as the way "to discard messages below a given priority" reproduces the class of defect this PR is fixing. The knob that actually works is MariaDB's own `log_warnings` (`sql/sys_vars.cc`: "Log some non-critical warnings to the error log. Value can be between 0 and 11. Higher values mean more verbosity", default `2`) — or filtering on the rsyslog side by message content.

## Nits

3. **"log a warning on every service start"** — it is logged at `LOG_NOTICE`, not warning (systemd's NEWS says "warnings", but the code is authoritative), and it is emitted when the unit file is *parsed*, so readers see it on `systemctl daemon-reload` and unit load rather than literally on every start.

4. **`StandardError=` defaults to `inherit`, not `journal`.** The effect is identical — it duplicates stdout's descriptor — but "This is already the default" is exact only for `StandardOutput=`. "already the effective default" would be right for both.

5. **"Only the forwarding setting differs from the defaults, so on a standard installation it is the only change needed"** overstates slightly. It assumes `log_error` is unset, which is why the first bullet exists — packaged installs often do set it. And the next sentence concedes that on `imjournal` distributions no change is needed at all. "usually the only unit-side change" would cover both.

6. **Sample log line.** The format string uses `%2d` for the hour, so `2026-01-01 0:00:00` would really render with two spaces. A two-digit hour sidesteps it — real output here reads `2025-11-19 12:50:03 0 [Note] …`.

Nothing above undermines the change; the obsolete-value fix and the `SyslogLevel=` analysis are both solid and were worth chasing past #963.

mariadb-stefan-hinz:
Thanks Caleb — folded all six points in as `3b00ad4bd`, so **please don't push your staged edits**, or we'll end up with two versions of the same change. Nothing else is needed from you.

What went in:

1. **`SyslogLevel=err` restored to the options table**, with the caveat in the Comments column and a link to the warning. The table is an equivalence table, and `mariadbd-safe --syslog` really does pipe output through `logger -p daemon.error`, so that row is the one carrying information.
2. **The `LogLevelMax=` recommendation is replaced.** It couldn't work: since MariaDB attaches no per-line priority, every line carries `SyslogLevel`, and `LogLevelMax=` filters on that same value — all or nothing, real errors included. Readers now get `log_warnings` for reducing what the server writes, or matching the `[Note]`/`[Warning]`/`[ERROR]` labels in the syslog daemon for sorting after the fact.
3. "log a warning on every service start" → a **notice**, logged when the unit file is **parsed** — so on `daemon-reload` and unit load too — and twice for a unit that sets both keys.
4. "already the default" → "already the **effective** default", since `StandardError=` defaults to `inherit`.
5. "the only change needed" softened: it assumes `log_error` is unset, and is moot where `rsyslog` already reads the journal.
6. Sample log line given a two-digit hour — the format string uses `%2d`, so `0:00:00` would really render with two spaces.

Your framing of the section is unchanged; this is the two corrections plus four nits. `doc-lint.sh` exit 0, `fragcheck` 33 fragment links / 0 dead.

One note on your `--syslog` point, which turned out to be a nice reinforcement rather than a contradiction: the "everything is tagged `err`, notes included" behaviour the warning objects to is exactly what `mariadbd-safe --syslog` has always done. So the hint isn't describing a hypothetical foot-gun — it's describing the thing the old advice was replicating.

PR #967, opened 2026-08-26
DOCS-6439 Restore superscripts in numeric size suffix table

mariadb-udaydubey:
The M/G/T/P/E rows of the size-suffix table in server-system-variables rendered as flat digits (10242, 10243, ...) because the exponent markup was lost in the original KB-to-GitBook migration. Restore them with <sup> so they read 1024^2 ... 1024^6.

Values verified against server mysys/my_getopt.c eval_num_suffix() (K=1<<10 ... E=1<<60).

Fix flattened superscript exponents in size-suffix table

The M/G/T/P/E rows in the numeric-size suffix table on server-system-variables.md rendered exponents as plain digits
(e.g. 10242 instead of 1024^2), lost during the original MariaDB-KB to GitBook migration. Restored using <sup> tags,
matching the existing table-cell superscript convention used elsewhere in the repo (CVE footnotes, upgrade-path markers),
and confirmed to survive GitBook sync via live precedent.

Values verified against mysys/my_getopt.c eval_num_suffix() (server @ bab03b0f).

### DIFF
--- a/server/ha-and-performance/optimization-and-tuning/system-variables/server-system-variables.md
+++ b/server/ha-and-performance/optimization-and-tuning/system-variables/server-system-variables.md
@@ -85,11 +85,11 @@ Variables that take a numeric size can either be specified in full, or with a su
 | Suffix | Description | Value                                                                                                                   |
 | ------ | ----------- | ----------------------------------------------------------------------------------------------------------------------- |
 | K      | kilobytes   | 1024                                                                                                                    |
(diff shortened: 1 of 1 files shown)


mariadb-TeemuOllakka (approved):
Looks good:
* Exponents are now rendered correctly
* `<sub>` is used also elsewhere in the docs, so it is a good choice here

PR #1000, opened 2026-09-01
DOCS-6520 Fix the isolation-level advice on the SQL Server page

mariadb-stefan-hinz:
[DOCS-6520](https://mariadbcorp.atlassian.net/browse/DOCS-6520). Follow-up to DOCS-6519 (#961), which corrected `SET TRANSACTION` and deliberately left this page alone.

## 1. The scalability claim is backwards

> Applications developed for SQL Server and later ported to MariaDB may run with `READ COMMITTED` without problems. **Using a stricter level would reduce scalability.**

Marko Mäkelä, in the thread that produced DOCS-6519:

> `READ COMMITTED` can be significantly slower than `REPEATABLE READ`, because it is creating a new read view at the start of each SQL statement. The read view creation was recently reworked in MDEV-21423 to reduce the impact.

The MDEV-21423 commit message says the same thing from the other side:

> Under high concurrency, MVCC snapshot creation may spend a significant amount of time in `lf_hash_iterate()`/`l_find()` while collecting active read-write transaction identifiers. This overhead is particularly visible in sysbench `oltp_read_write` with `transaction-isolation=READ-COMMITTED`.

That commit (`241050cca82`) is on `main` and is tagged `mariadb-11.8.9` and `mariadb-12.3.3` — so it is a mid-series improvement, and older series do not have it. The page now says READ COMMITTED is not a throughput setting, and points at a new section rather than repeating the argument.

## 2. The config-file recipe stops the server from starting

`tx_isolation` is `NO_CMD_LINE` (`sql/sys_vars.cc:4658`); the startup option is `transaction-isolation` (`sql/mysqld.cc:7287`), whose accepted values are the hyphenated `tx_isolation_names` (`sql/handler.cc:156`). Confirmed against a 12.3.3 binary:

```
--transaction-isolation=name
                    Default transaction isolation level. One of: READ-UNCOMMITTED,
                    READ-COMMITTED, REPEATABLE-READ, SERIALIZABLE
```

`tx_isolation` does not appear in `--verbose --help` at all. The recipe is now `transaction-isolation = READ-COMMITTED`, with a sentence saying why the other spelling fails.

## 3. `tx_isolation` deprecated since 11.1

`DEPRECATED(1101, "transaction_isolation")` at `sql/sys_vars.cc:4661`. Live on 12.3.3:

```
Warning 1287  '@@tx_isolation' is deprecated and will be removed in a future
release. Please use '@@transaction_isolation' instead
```

All three uses on the page move to `transaction_isolation`, and the link target moves to `#transaction_isolation`.

## 4. Two locking claims corrected

*"In MariaDB, the locks acquired by a read do not depend on the isolation level (with one exception noted below)."* They do — `storage/innobase/row/row0sel.cc:4776`:

```c
/* Do not lock gaps at READ UNCOMMITTED or READ COMMITTED
isolation level */
const bool set_also_gap_locks =
        prebuilt->select_lock_type != LOCK_NONE
        && trx->isolation_level > TRX_ISO_READ_COMMITTED
```

The promised "one exception noted below" was never actually named anywhere on the page. Replaced with the real dependency, plus the `SERIALIZABLE` case from `ha_innodb.cc:16811`, where a plain `SELECT` in a non-autocommit transaction becomes `LOCK IN SHARE MODE`.

*"`REPEATABLE READ` does not acquire share locks on all read rows, nor a range lock on the missing values that match a `WHERE` clause."* True of plain reads only — the same guard above makes a **locking** read at `REPEATABLE READ` take next-key locks. Now scoped.

## New section on SET TRANSACTION

The question that started all of this — *"for high volume systems, does MariaDB recommend setting the isolation level to READ COMMITTED?"* — had no answer anywhere in the docs. `SET TRANSACTION` now has a **Choosing an Isolation Level** section that answers it: stay on the default, and here is what READ COMMITTED buys and costs.

## Drive-by fixes

- `server-system-variables.md`: `tx_isolation` and `tx_read_only` both listed a command line of their own. Both are `NO_CMD_LINE` (`sys_vars.cc:4658`, `:4719`); the options belong to the replacement variables. Reworded rather than removed, since the option does reach the same storage.
- Migration page: link text `SELECT ... LOCK IN SHARED MODE` → `LOCK IN SHARE MODE` (the statement's actual name; the link target was already right).
- Migration page: two `sql` snippets were fenced as `bash`.
- Migration page: `tx_isolation` in the frontmatter description.

## Checks

`doc-lint.sh` clean on all three files, CI-exact codespell clean, `fragcheck.py new main` reports no new dead anchors, and every added relative link resolves against the tree.

[DOCS-6520]: https://mariadbcorp.atlassian.net/browse/DOCS-6520?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/server/reference/sql-statements/transactions/set-transaction.md
+++ b/server/reference/sql-statements/transactions/set-transaction.md
@@ -157,6 +157,10 @@ Distributed [XA transactions](xa-transactions.md) should always use this isolati
 
 If the [innodb\_snapshot\_isolation](../../../server-usage/storage-engines/innodb/innodb-system-variables.md#innodb_snapshot_isolation) system variable is not set to `ON`, strictly speaking anything other than `READ UNCOMMITTED` is not clearly defined. While it is `ON`, an attempt to acquire a lock on a record that does not exist in the current read view raises an error and rolls the transaction back.
 
+Its default is not the same across supported releases. It is `ON` from MariaDB 11.6.2, and `OFF` in the earlier series that introduced it, which includes the 10.11 and 11.4 long-term releases. Advice to keep the default is therefore ambiguous; set the variable explicitly.
+
+Prefer `ON`, because it is what makes `REPEATABLE READ` behave as its name claims. The reason to choose `OFF` is application compatibility, not performance. With `ON`, an `UPDATE` or `DELETE` that touches a row changed since the transaction's snapshot fails with [ER\_CHECKREAD](../../error-codes/mariadb-error-codes-1000-to-1099/e1020.md) (1020) and the whole transaction is rolled back, exactly as for a deadlock; the server treats the two errors as the same class, and replication retries both. An application that already retries on `ER_LOCK_DEADLOCK` needs no change. A legacy application that does not handle `ER_CHECKREAD` the same way loses those transactions instead of retrying them, and `innodb_snapshot_isolation=OFF` gives it the MySQL-compatible — though not well-defined — behavior until it can be fixed.
+
 ### Access Mode
 
 {% tabs %}
@@ -175,7 +179,14 @@ It is not permitted to specify both `READ WRITE` and `READ ONLY` in the same sta
 
 ## Choosing an Isolation Level
 
-Stay on the default `REPEATABLE READ` unless the application needs the semantics of a different level. In particular, `READ COMMITTED` is not a throughput setting: it opens a new read view at the start of every statement, and under high concurrency the cost of building those snapshots can make it slower than `REPEATABLE READ`, not faster. That path was made cheaper by [MDEV-21423](https://jira.mariadb.org/browse/MDEV-21423) from MariaDB 11.8.9 and 12.3.3, which reduces the overhead without removing it.
+Stay on the default `REPEATABLE READ` unless the application needs the semantics of a different level. `READ COMMITTED` is not a throughput setting in either direction: which level costs more depends on the shape of the workload, because the two differ in *when* a read view is opened.
+
+At `READ COMMITTED`, every statement opens its own read view. At `REPEATABLE READ`, all statements share the read view taken when the transaction started. That single difference cuts both ways:
+
+* **Many short statements in a recently started transaction are more expensive at `READ COMMITTED`.** Opening a read view means collecting the identifiers of the read-write transactions that are currently active, and under high concurrency that collection is measurable. This is a common shape for web applications. [MDEV-21423](https://jira.mariadb.org/browse/MDEV-21423) reduced the cost from MariaDB 11.8.9 and 12.3.3 without removing it.
+* **Complex statements in a long-running transaction can be cheaper at `READ COMMITTED`.** Each statement is allowed to see more recent data, so rebuilding the row versions it needs walks less history. At `REPEATABLE READ` the read view stays fixed at the start of the transaction, and the older it gets, the further back through the undo history InnoDB has to walk to reconstruct the snapshot that a statement is entitled to see.
+
+So measure the workload rather than assume a direction.
 
 The reasons to choose `READ COMMITTED` are about behavior rather than speed:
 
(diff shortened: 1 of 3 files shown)


server/reference/sql-statements/transactions/set-transaction.md
```diff
@@ -173,6 +173,22 @@ It is not permitted to specify both `READ WRITE` and `READ ONLY` in the same sta
 
 `READ WRITE` and `READ ONLY` can also be specified in the [START TRANSACTION](start-transaction.md) statement, in which case the specified mode is only valid for one transaction.
 
+## Choosing an Isolation Level
+
+Stay on the default `REPEATABLE READ` unless the application needs the semantics of a different level. In particular, `READ COMMITTED` is not a throughput setting: it opens a new read view at the start of every statement, and under high concurrency the cost of building those snapshots can make it slower than `REPEATABLE READ`, not faster. That path was made cheaper by [MDEV-21423](https://jira.mariadb.org/browse/MDEV-21423) from MariaDB 11.8.9 and 12.3.3, which reduces the overhead without removing it.
```
dr-m: It is not this straightforward.

Yes, `READ COMMITTED` can be more expensive than `REPEATABLE READ` when running many simple statements in a single transaction that was started recently. This could be common in web applications.

However, when the workload includes complex statements that are executed in a long-running transaction, resetting the read view more frequently (`READ COMMITTED`) means that the statements are allowed to see more recent data. With `REPEATABLE READ`, all statements would use the same read view and might have to traverse through more history to reconstruct the snapshot that corresponds to the start of the transaction, instead of the start of the current statement.

server/reference/sql-statements/transactions/set-transaction.md
```diff
@@ -173,6 +173,22 @@ It is not permitted to specify both `READ WRITE` and `READ ONLY` in the same sta
...
+* InnoDB takes no gap locks, so an insert into a range that another transaction has locked is not blocked. See [Gap Locking at READ COMMITTED](#gap-locking-at-read-committed). On its own this is not a reason to expect better throughput, for the read view reason above.
+
+What you give up in exchange:
+
+* Reads are no longer repeatable. Two identical `SELECT` statements in the same transaction can return different rows.
+* Statement-based binary logging is not available, as described under [Gap Locking at READ COMMITTED](#gap-locking-at-read-committed).
+
+Whichever level you choose, leave [innodb\_snapshot\_isolation](../../../server-usage/storage-engines/innodb/innodb-system-variables.md#innodb_snapshot_isolation) at its default. See [innodb\_snapshot\_isolation](#innodb_snapshot_isolation) above.
```
dr-m: This advice is quite confusing, because the default was only recently changed, in [MDEV-35124](https://jira.mariadb.org/browse/MDEV-35124), MariaDB Server 11.8. The 10.11 and 11.4 releases are still supported.

I would suggest to reword this. There could be reasons for or against enabling this setting. If you run legacy applications that do not handle the error code `ER_CHECKREAD` in a similar way as `ER_DEADLOCK`, you might want to use the MySQL compatible and academically invalid setting `innodb_snapshot_isolation=OFF`.

server/reference/sql-statements/transactions/set-transaction.md
mariadb-stefan-hinz: Thank you — you are right, and the page was stating only one half of it. Rewritten so the trade-off cuts both ways, with your long-transaction case spelled out: at `REPEATABLE READ` the read view is pinned at transaction start, so the older it gets the further back InnoDB walks the undo history to rebuild the version a statement is entitled to see, whereas `READ COMMITTED` lets each statement reconstruct from a nearer point.

Checked against source before rewording: `row_vers_build_for_consistent_read()` loops on `trx_undo_prev_version_build()` until a version is visible to the view (`row0vers.cc:704`), and `ha_innodb.cc:17082` closes the read view per statement at `READ COMMITTED` / `READ UNCOMMITTED`. The MDEV-21423 sentence is kept but now sits under the short-statement bullet only, where it belongs, and the section closes on measuring rather than on a direction.

The SQL Server migration page carried the same one-directional claim and is corrected the same way.

server/reference/sql-statements/transactions/set-transaction.md
mariadb-stefan-hinz: Agreed, and reworded. "Leave it at its default" was the wrong advice precisely because the default is not one thing across supported releases — that is the confusion you are pointing at, so the page now says to set it explicitly and gives the reasons either way.

One detail for the record: MDEV-35124 (`4e1e9ea6f32`) landed a little earlier than 11.8 — its earliest tag is `mariadb-11.6.2`, so the split is ON from 11.6.2 and OFF in the series that introduced the variable (`b8a67198895`), which is where 10.11 and 11.4 sit. Your point stands either way: both of those LTS lines default to OFF and are supported.

Your `ER_CHECKREAD` framing is now the stated reason to choose OFF — application compatibility, not performance. The page says the error rolls back the whole transaction exactly as a deadlock does, so an application that already retries on `ER_LOCK_DEADLOCK` needs no change and one that does not will lose transactions until it is fixed. Source for the "same class" claim: InnoDB handles `DB_RECORD_CHANGED` and `DB_DEADLOCK` in the same case arm (`ha_innodb.cc:2160`, `row0mysql.cc:711`), and replication lists `ER_CHECKREAD` among its retryable errors next to `ER_LOCK_DEADLOCK` (`slave.cc:1015`). "MySQL compatible and academically invalid" is in there too, as "the MySQL-compatible — though not well-defined — behavior".

PR #1002, opened 2026-09-01
DOCS-6539 Add missing Galera provider options to wsrep_provider_options

mariadb-udaydubey:
## What

Documents six Galera provider options that are registered in the provider but were missing from
the `wsrep_provider_options` reference page. Follows up on DOCS-6506 (PR #978), which added
`pc.bootstrap` / `socket.ssl_reload` and asked for the page's full option list to be diffed
against the provider's registry.

| Option | Dynamic | Default |
|--------|---------|---------|
| `gcache.keep_plaintext_size` | Yes | value of `gcache.page_size` |
| `gcs.check_appl_proto` | Yes | `1` |
| `gcs.stateless` | No | `false` |
| `gcs.vote_policy` | No | `0` |
| `gmcast.group` | No | None (internal, set from cluster name) |
| `gmcast.peer_addr` | Yes | None |

## Scope — 6 documented, 4 excluded

The ticket named ten options. Source verification narrowed it to six; the other four are excluded
on the ticket's **own** criteria:

- **`gcache.debug`** — registered only in debug builds (`#ifndef NDEBUG`), same class as the
  ticket's already-excluded `dbug` / `signal` / `gcs.sm_dump`.
- **`socket.non_blocking`, `gmcast.mcast_port`, `gmcast.mira`** — all carry the provider's
  `hidden` flag (`gcomm/src/defaults.hpp`), two with the literal source comment
  *"Hidden because undocumented"* — the same rationale that keeps `ist.keep_keys` off the page.

(`gmcast.isolate` is also `hidden` yet already documented, so `hidden` isn't an airtight bright
line in the source — hence this was a deliberate editorial call, flagged for review.)

## Verification

Source-verified against **MariaDB/galera @ `4cadb3c2ce3eaf95e5baf74c97b216414a77b803`**
(tag `mariadb-26.4.28`). Dynamic-ness was read from each option's `set_param`/`param_set` handler
rather than inferred from the `read_only` flag; defaults confirmed at each registration site. Full
claim-by-claim fact-check report is posted on the Jira ticket.

Two descriptions (`gcs.check_appl_proto`, `gcache.keep_plaintext_size`) are name-derived — their
default and dynamic-ness are source-proven, but the source gives no prose semantics, so the
Description wording is a best reading and marked UNVERIFIED in the fact-check report. Confirmation
from the Galera team welcome.

## Ticket

DOCS-6539

### DIFF
--- a/galera-cluster/reference/wsrep-variable-details/wsrep_provider_options.md
+++ b/galera-cluster/reference/wsrep-variable-details/wsrep_provider_options.md
@@ -181,6 +181,12 @@ Note that before Galera 3, the `repl` tag was named `replicator`.
 * Dynamic: No
 * Default: `0`
 
+#### `gcache.keep_plaintext_size`
+
+* Description: Encryption-related. Amount of GCache data kept in plaintext.
+* Dynamic: Yes
+* Default: The value of [gcache.page\_size](wsrep_provider_options.md#gcache.page_size).
+
 #### `gcache.mem_size`
 
 * Description: Maximum size of size of the malloc() store for setups that have spare RAM.
@@ -249,6 +255,12 @@ Note that before Galera 3, the `repl` tag was named `replicator`.
 * Dynamic: No
 * Default: Empty string
 
+#### `gcs.check_appl_proto`
+
+* Description: Controls whether the application protocol version is checked.
+* Dynamic: Yes
+* Default: `1`
+
 #### `gcs.fc_debug`
 
 * Description: If set to a value greater than zero (the default), debug statistics about SST flow control will be posted each timegcs.fc\_master\_slave after the specified number of writesets.
@@ -306,11 +318,29 @@ Note that before Galera 3, the `repl` tag was named `replicator`.
 * Dynamic: No
 * Default: `0.25`
 
+#### `gcs.stateless`
+
+* Description: For internal use. Should not be manually set.
+* Dynamic: No
+* Default: `false`
+
 #### `gcs.sync_donor`
 
 * Description: Whether or not the rest of the cluster should stay in sync with the donor. If set to `YES` (`NO` is default), if the donor is blocked by state transfer, the whole cluster is also blocked.
 * Dynamic: No
 * Default: `no`
+
+#### `gcs.vote_policy`
+
+* Description: For internal use. Should not be manually set. Determines the cluster's voting policy, which must be decided before the cluster starts and cannot be changed at runtime.
+* Dynamic: No
+* Default: `0`
+
+#### `gmcast.group`
+
+* Description: For internal use. Should not be manually set. Set by the provider from the cluster name.
+* Dynamic: No
+* Default: None
 
 #### `gmcast.isolate`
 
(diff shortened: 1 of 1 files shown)


mariadb-udaydubey:
Hey @seppo

PR is updated and ready for a final look.

Summary of changes since your first review:

All six description rewrites from your review are applied (gcs.stateless, gcs.vote_policy, gcs.check_appl_proto, gcache.keep_plaintext_size, gmcast.group, gmcast.peer_addr).
Following your suggestion, Teemu weighed in on gmcast.group and gmcast.mira as well. Net result:
gmcast.group — removed from public docs (Teemu confirmed the value is unconditionally overridden by wsrep_cluster_name, so a user-set value is silently discarded — source-traced and confirmed).
gmcast.mira — removed (Teemu: reportedly unused in practice).
gmcast.mcast_port — added, per your and Teemu's input.
Final scope: 6 options documented, 4 excluded (gcache.debug, socket.non_blocking, gmcast.group, gmcast.mira).

One open item I'd appreciate your eyes on: gcs.vote_policy's description presents 1 as a special case nested inside the N > 0 threshold rule, based on my best reading of your original comment (which arrived with some formatting mangled). Wanted to confirm that's the correct structure rather than a fourth, standalone case — let me know if I read that wrong.

Full source citations for everything are in the attached fact-check report. Thanks again for the thorough review — genuinely improved the accuracy here.

galera-cluster/reference/wsrep-variable-details/wsrep_provider_options.md
```diff
@@ -338,12 +366,27 @@ Note that before Galera 3, the `repl` tag was named `replicator`.
 * Dynamic: No
 * Default: None
 
+#### `gmcast.mcast_port`
+
+* Description: The UDP port used by GMCast's optional IP multicast transport. It is only consulted when multicast is enabled by setting [gmcast.mcast\_addr](wsrep_provider_options.md#gmcast.mcast_addr). Multicast is disabled by default, since that option is empty. When it is not set, the multicast group uses the GMCast listen port. Set it only when the multicast group has to use a port other than `4567`. The option carries the provider's `hidden` flag, so it does not appear in the [wsrep\_provider\_options](../galera-cluster-system-variables.md#wsrep_provider_options) output.
+* Dynamic: No
```
sjaakola, line 372: I suggest to not mention the provider hidden flag for the end users, and leave the last sentence out.

sjaakola (changes requested):

galera-cluster/reference/wsrep-variable-details/wsrep_provider_options.md
mariadb-udaydubey: @sjaakola Makes sense. That's more of an internal implementation detail than something a reader needs to reason about. Will remove that sentence now.

While I have your attention: is there anything else across the six documented options (or the exclusions) you'd want me to double-check or adjust? Would rather catch everything in one pass. Happy to update it all together and get this closed out.

TIA.

sjaakola: the other option documentations are ok, no other change requirements from me

mariadb-udaydubey: @sjaakola I removed the `hidden flag` sentence. You can review it and after that I close the ticket.

sjaakola (approved):
looks good now

PR #1004, opened 2026-09-01
DOCS-6567 Fail the lint when a page is missing from SUMMARY.md

dbart:
Fixes [DOCS-6567](https://mariadbcorp.atlassian.net/browse/DOCS-6567). Follow-up to DOCS-6566 (#1003).

## The gap

GitBook publishes only the pages listed in a space's `SUMMARY.md`. A page file with no nav entry never renders — and **no gate in this repo could see that**. The markup is valid, so codespell passes. The links resolve, so lychee passes. The page is simply never built, so there is no failing signal anywhere. The only symptom is a reader reporting a missing page.

That is how DOCS-6566 happened: `dde0fb263` added four post-download pages and bumped their `most-recent-*.md` includes but never touched `platform/SUMMARY.md`. All four sat unpublished for eight days, every check green throughout, until a reader noticed.

Before this PR the repo's only orphan check was check 6 of the `docs-check` skill — an LLM-performed heuristic that runs solely when a human invokes it.

## The change

New `.claude/hooks/navcheck.py`, wired into `doc-lint.sh` as a fourth check with no CI counterpart.

**It is history-aware, and that is the load-bearing design decision.** `main` carries **219** pre-existing orphans (190 in `server` alone, measured across the 11 spaces that have a `SUMMARY.md`: 9,468 `.md` files vs 9,238 nav refs). An absolute check would fail every unrelated PR on breakage it did not introduce — precisely the trap already documented for the heading-anchor gate and its ~1,272 pre-existing dead anchors. So it diffs against `DOC_LINT_BASE` and reports only pages *newly* orphaned:

- a page added with no nav entry — the DOCS-6566 case;
- a nav entry deleted while the page file survives — the quieter direction, since the page keeps working locally and only vanishes from the built site.

**Unlike `fragcheck.py` it needs no worktree.** It only needs a file list and one file's contents per space, so the base side is read with `git ls-tree` and `git show`. That is why it costs ~40 ms scoped to a space (209 ms across all 11) against fragcheck's ~14 s, and why it has no skip flag.

## Verification

Replaying the real `dde0fb263` in a worktree at that commit:

```
navcheck: 4 page(s) added or de-listed but not listed in SUMMARY.md:
  platform/post-download/mariadb-server-10.11.19.md
  platform/post-download/mariadb-server-11.4.13.md
  platform/post-download/mariadb-server-11.8.9.md
  platform/post-download/mariadb-server-12.3.3.md
exit=1
```

It catches the actual incident and names all four pages.

| Scenario | Expected | Actual |
| --- | --- | --- |
| clean tree, all spaces | 0 | 0 |
| added page with no nav entry | 1 | 1 |
| …acknowledged via `DOC_LINT_ALLOW_ORPHAN` | 0 | 0 |
| nav entry deleted, page survives | 1 | 1 |
| edit to a **pre-existing** orphan | 0 | 0 |
| unresolvable base rev → SKIP | 0 | 0 |
| file outside any space | 0 | 0 |
| bad mode → usage error | 2 | 2 |
| `doc-lint.sh` end-to-end, clean tree | 0 | 0 |

`py_compile`, `bash -n` and `shellcheck -S warning` are clean; `doc-lint.sh` passes on the changed docs.

## Scope decisions

- A directory counts as a space only if it holds a `SUMMARY.md`. That keeps `dev-docs/`, `.claude/` and `help-tables/` out without hardcoding names.
- `.gitbook/` is excluded — 117 include and reusable-snippet files that are pulled in by `{% include %}` and are never nav-listed.
- The space's own `SUMMARY.md` is not a page in itself.
- Cross-space `SUMMARY` entries are absolute `app.gitbook.com` URLs (6 today) and are skipped; they point into another space and can never mark a local file as referenced.
- `DOC_LINT_ALLOW_ORPHAN` (space/comma-separated, or `all`) exempts a deliberately unlisted page, matching how `DOC_LINT_ALLOW_SHRINK` works for the gutted-page guard.

## Docs

`dev-docs/cookbook-pre-pr.md`, `.claude/skills/docs-check/SKILL.md` and `.claude/README.md` are updated. Check 6 of the skill is rewritten: its orphan half is now automated, and the note says plainly that being an invoke-only heuristic is how DOCS-6566 shipped. What remains there is the dangling direction (nav entries pointing at deleted files) and whether a moved page's entry sits in the right place — both genuinely editorial.

## Note for reviewers

**CI does not exercise this.** The gate runs in the pre-commit hook and via `docs-check`, not in a workflow, so the green checks on this PR say nothing about it. The replay above is the evidence. Adding a CI job is a separate change and deliberately not bundled here.

Triaging the 219 pre-existing orphans is also out of scope — some are surely intentional, some are likely lost content of exactly the DOCS-6566 kind. `.claude/hooks/navcheck.py check` prints the full inventory for whoever picks that up.

[DOCS-6567]: https://mariadbcorp.atlassian.net/browse/DOCS-6567?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/.claude/README.md
+++ b/.claude/README.md
@@ -7,7 +7,9 @@ Code. It contains:
 |------|------------|
 | `settings.json` | Project settings, incl. the `PreToolUse(Bash)` hook wiring |
 | `settings.local.json` | **Personal** overrides — gitignored, never committed |
-| `hooks/doc-lint.sh` | Canonical codespell + lychee linter (single source of truth, mirrors CI) |
+| `hooks/doc-lint.sh` | Canonical codespell + lychee linter (single source of truth, mirrors CI), plus four checks with no CI counterpart: includes, heading anchors, orphaned pages, gutted pages |
+| `hooks/fragcheck.py` | GitBook-accurate heading-anchor checker, called by `doc-lint.sh` |
+| `hooks/navcheck.py` | Orphaned-page (nav coverage) checker, called by `doc-lint.sh` |
 | `hooks/pre-commit.sh` | PreToolUse hook: gates Claude-made `git commit`s by calling `doc-lint.sh` |
 | `skills/` | Shared skills (e.g. `docs-check`) |
 | `commands/` | Shared slash commands (e.g. `/precommit`) |
--- a/.claude/hooks/doc-lint.sh
+++ b/.claude/hooks/doc-lint.sh
@@ -1,9 +1,9 @@
 #!/usr/bin/env bash
 #
 # doc-lint.sh — the SINGLE SOURCE OF TRUTH for the codespell + lychee invocations that mirror
-# CI (.github/workflows/codespell.yml and link-check-pr.yml), plus three checks that have no CI
-# counterpart (see below): a GitBook include resolver, a heading-anchor gate, and a net
-# line-loss guard.
+# CI (.github/workflows/codespell.yml and link-check-pr.yml), plus four checks that have no CI
+# counterpart (see below): a GitBook include resolver, a heading-anchor gate, an orphaned-page
+# (nav coverage) gate, and a net line-loss guard.
 #
 # The pre-commit hook, the /precommit command, the docs-check skill, and dev-docs/cookbook-pre-pr.md
 # all delegate here instead of re-spelling the flags, so the CI-mirroring options live in exactly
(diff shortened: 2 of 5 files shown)


mariadb-stefan-hinz:
Looks good to me. Merge it, I'd say!

PR #1006, opened 2026-09-02
DOCS-6508 Fix four nonexistent variables on server reference pages

mariadb-udaydubey:
Four variables were documented on server reference pages but exist in **neither** the Community (13.1.0) nor Enterprise (10.6.28) server source — verified by exact-string `git grep` against both trees (fact-check report on DOCS-6508).

- **`legacy_xa_rollback_at_disconnect`** — 0 hits; no XA rollback-at-disconnect variable exists. Removed the fabricated work-around sentence in `xa-transactions.md` and the entry in the full options/variables list.
- **`show_old_temporals`** — MySQL-only (present only in a mysql-test file as "Port … from MySQL 5.6"). Removed the `--show-old-temporals` mariadbd option and the `show_old_temporals` system-variable list entry.
- **`spider_reset_auto_increment`** — no such Spider variable. The sentence sits inside `spider_auto_increment_mode`'s own mode-3 description, so renamed to `spider_auto_increment_mode`.
- **`write_buffer_size`** — a myisamchk client option, not a server system variable; annotated as such on the myisamchk memory page.

**Spider rename — partially confirmed:** A Spider engineer confirmed `spider_auto_increment_mode` is the likely intended variable (informal review, not a full sign-off). They were unable to confirm whether the accompanying behavioral claim — that modes 2/3 are what you'd use "to use an auto-increment column on the remote server" — is accurate, due to time constraints. The rename is included in this PR; the behavioral description is pre-existing doc text, carried over unchanged, and has not been independently verified by this fix.

Ticket: https://mariadbcorp.atlassian.net/browse/DOCS-6508


mariadb-stefan-hinz (approved):
Looks good to me. Merge it! :-)

PR #1007, opened 2026-09-02
mariadb R2DBC 1.4.2 release notes

rusher:
Here is the documentation for connector R2DBC 1.4.2 release


dbart:
Thanks Diego — before anything else: **the red `expand-links` check is not your content.** It's a fork limitation in the workflow, and there's nothing for you to fix.

Step by step, from the [job](https://github.com/mariadb-corporation/mariadb-docs/actions/runs/33642022266/job/100287207435):

| # | Step | Result |
|---|------|--------|
| 3 | Expand GitBook aliases (Stable Version) | **success** |
| 4 | Commit and push changes back to PR branch | **failure** — exit 128 |
| 5 | Fail on an unexpanded alias | **skipped** |

The expansion itself worked. What failed is step 4 pushing the expanded result back to the branch: this PR comes from `rusher/mariadb-docs`, and on a `pull_request` event the workflow's `GITHUB_TOKEN` has no write access to a fork's branch — hence the git exit 128. Step 5 is the gate that would actually complain about a bad alias, and it never ran, so nothing here says your `{connectors}` links are wrong.

Corroborating: every other recent run of that workflow is green, and `main-r2dbc` is the only fork branch among them.

`maintainerCanModify` is true on this PR, so I can run the expansion and push it to your branch from this side — that's the simplest path and I'll take care of it. No action needed from you on CI.

The rest of the review is separate and I'll follow up shortly; the release-notes side looks good so far — SUMMARY has both the page and changelog entries in the right places, the `latest-r2dbc` include is updated in both spots, and the frontmatter and body dates agree. Also noticed your two edits to `1.4.1.md` are real fixes rather than collateral: it was the only R2DBC release-note page missing the `latest-r2dbc` include, and `../../` is indeed the correct depth for `release-criteria.md` from that directory. Nice catches.

dbart (approved):

dbart:
Expansion pushed as `ae661ec34`, and **CI is fully green now** — including `expand-links`.

Worth knowing why, since it's the useful takeaway for future connector PRs from a fork: the workflow's push-back step only runs when the expansion actually changed something. With the aliases already expanded in the branch, its `git diff --quiet` check finds nothing, prints "No alias replacements needed", and exits 0 — so the step that couldn't push never needs to, and the "Fail on an unexpanded alias" gate runs and passes. **Expanding locally before pushing sidesteps the fork limitation entirely.** Nothing was wrong with your aliases; all four `{connectors}` targets were valid names.

I also pushed a second commit, `dfa37c16f`, extending the fix you made on `1.4.1.md` to four older pages: `1.2.3`, `1.3.0`, `1.3.1` and `1.4.0` all still linked to `../../../community-server/about/release-criteria.md`. Three levels up from `release-notes/connectors/r2dbc/` leaves the release-notes space and lands on a repo-root `community-server/` that doesn't exist, so those were broken links. `../../` — your depth — is correct, and all four now resolve. They'd never been caught because lychee only lints the files a PR changes, so a page nobody touches keeps a broken link indefinitely. Your 1.4.1 fix is what surfaced the pattern.

The rest of the review checks out, so from my side there's nothing outstanding for you:

- `SUMMARY.md` has entries for both the page and the changelog, in the right nav positions.
- `latest-r2dbc.md` is updated in both places — the version link and the Download button label. That second one is easy to miss.
- The frontmatter and body release dates agree (`2026-09-02` / `2 Sep 2026`). For what it's worth, `1.4.1` disagrees with itself there; good that the new page doesn't inherit it.
- Adding the `latest-r2dbc` include to `1.4.1.md` was right — it was the only R2DBC release-note page missing it, 14 of the other 15 have it.
- The `maxAllowedPacket` row in the connection-parameters table matches R2DBC-128 and is correctly marked "Since 1.4.2"; the guide and skill version bumps are consistent, including the `0.9.1-spec` coordinates; and the two `</dependency>` newline fixes in the quickstart guide are a nice cleanup while you were in there.

Thanks for the thorough PR — the include and nav updates being right first time is the part that usually gets missed. Merge is on me from here.

PR #1009, opened 2026-09-02
DOCS-6570 Document what GitBook does when a published page is renamed

dbart:
[DOCS-6570](https://mariadbcorp.atlassian.net/browse/DOCS-6570) — `dev-docs/cookbook-gitbook-redirects.md`, +45 lines, no deletions. Agent/contributor playbook only; no published page is touched.

## Why

The cookbook covered the CSV format and the Cloudflare failure modes in detail, but said nothing about how GitBook resolves a page that has been **renamed** — which is the most common reason to reach for redirects in the first place. Three behaviors decide whether a rename needs redirects at all, and none is what you would guess from the existing text.

I know they aren't, because DOCS-6408 got the first one wrong. That PR predicted two stored redirect rules would shadow three renamed pages; the live site disproved it the moment Git Sync landed. The prediction was reasonable given what the cookbook said, which is the argument for writing this down rather than just knowing it.

## What the new section says

Established empirically on DOCS-6408, which renamed three published `platform/post-download/` pages:

1. **A real page outranks a redirect rule with the same source.** GitBook resolves an existing page first and falls through to a rule only when nothing resolves — so a rename cannot be shadowed by a rule.
2. **Git Sync preserves a page's id across a rename**, and rule destinations are page ids rather than paths (`destination.kind: site-page`). A rule pointing at a renamed page therefore follows it, which is convenient — but a rule *sourced from* the new canonical path becomes a **self-redirect**. That is the loop the Troubleshooting section already calls worse than a 404 and invisible to link checkers, reachable without any help from Cloudflare, and latent for as long as the page keeps winning.
3. **The old URL keeps working with no rule at all**, via an automatic slug-history alias. On DOCS-6408 two of three old URLs redirected correctly with nothing stored, so the prepared CSV had nothing left to import.

Plus a short follow-on procedure — audit rules sourced from the **new** canonical path (the old path is the obvious half; the new path is where a self-redirect hides), probe before preparing a CSV, delete self-redirects only *after* Git Sync publishes, and note that deletion goes through `bulkUpsertSiteRedirects` with a null destination since the API has no single-rule delete.

Two checklist items, and a back-link from the self-redirect failure mode so it no longer reads as purely Cloudflare's doing.

## Checks

`.claude/hooks/doc-lint.sh` exits 0. Both internal anchors resolve to real headings.

## One thing to push back on if you disagree

"Slug-history alias" is my name for the third mechanism, not GitBook's — the API does not expose it. What is proven is the observable: `nodejs-3.5.0` and `nodejs-3.5.1` redirected to the canonical path with no rule sourced from either, confirmed against the rule listing before any were created. The section claims the observable behavior, not the internals. If you would rather it read "GitBook keeps the old URL working on its own" without naming a mechanism, that is a one-line change.

## Not in this PR

The `gitbook-redirects` skill states redirects have "no API/MCP" path and its guardrail forbids using one, while this cookbook documents that path — and DOCS-6408 used it for both reads and writes. The skill and the cookbook now disagree more visibly than before. That belongs in a skill-bug ticket against the skill (owner: you), not in a cookbook edit.

[DOCS-6570]: https://mariadbcorp.atlassian.net/browse/DOCS-6570?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/dev-docs/cookbook-gitbook-redirects.md
+++ b/dev-docs/cookbook-gitbook-redirects.md
@@ -13,6 +13,8 @@ GitBook itself **does** support redirects, configured two ways:
   `getSiteRedirectBySource`, `createSiteRedirect`, `updateSiteRedirectById`,
   `bulkUpsertSiteRedirects`. Useful for bulk loads and, above all, for **verifying** what is
   actually configured — see [Troubleshooting](#troubleshooting-a-redirect-that-looks-broken).
+  **Requires `admin` on the site**, which the organization's default `create` role does not
+  grant — so for most writers the CSV above is the only route, and that is by design.
 
 ## When to produce a redirect CSV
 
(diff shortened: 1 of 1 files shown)


mariadb-stefan-hinz (approved):
Looks like a good addition to the cookbook. Merge it!

PR #1012, opened 2026-09-03
MDBF-1241: Automation of generation and PR submission for plugin API docs

gkodinov:
Implemented a daily runner github action that will:

* fetch the server repository
* generate the plugin API markdown docs from it
* install the markdown docs into mariadb-docs/server/reference/plugins/plugin-api/
* open a pull request to mariadb-docs if the generated docs are changed

Co-authored-by: Grok


mariadb-stefan-hinz:
I reviewed. Looks good to me. A few comments/suggestions, though:

- `server/reference/plugins/plugin-api/` is where the docs will end up. I suggest to make that `api-plugin` , so that it's more in line with the other plugins (like e.g. `server/reference/plugins/password-validation-plugins`).
- The script doesn't write `SUMMARY.md` entries. That's fine, since I prefer to write those entries myself and verify they look right. (Without them, pages would live on GitHub but not be published on [mariadb.com/docs.](http://mariadb.com/docs.)) However, if the set of pages changes (pages added, removed, or renamed slug-wise), entries in `SUMMARY.md` need to change accordingly. The reviewer of the PRs created by the script needs some kind of heads-up about that.

The SUMMARY.md entries need to land here: https://github.com/mariadb-corporation/mariadb-docs/blob/main/server/SUMMARY.md?plain=1#L2188

mariadb-stefan-hinz:
Reviewed at Stefan's request, as a second pass on top of his 2026-09-04 comment. The approach is right — generating from the server tree and letting a bot open the PR is the correct shape for this content, and the pieces that worry me most in an automation PR (what it reads, and what it deletes) both check out. Three blockers below, then correctness notes.

## Blockers

**1. There are two workflows now, and both are scheduled.**

`sync-server-plugin-api-docs.yml` and `sync-server-api-plugin-docs.yml` are near-identical: same `name:`, same `cron: '17 5 * * *'`, same generation step. They differ only in destination — `server/reference/plugins/plugin-api/` vs `api-plugin/` — and the `api-plugin` one additionally runs the SUMMARY script. Their `concurrency` groups differ, so they will *not* serialize against each other: every morning at 05:17 UTC both fire, install the same generated content into two different directories, and open two PRs from two branches. In the Actions UI they show up as two rows with the same display name.

This reads like Stefan's `plugin-api` → `api-plugin` rename being added as a new file instead of an edit to the old one. Deleting `sync-server-plugin-api-docs.yml` should be all that's needed. While you're there: the surviving workflow's PR body still says *"Automated by `.github/workflows/sync-server-plugin-api-docs.yml`"* — the other filename.

**2. The script isn't executable, so the step fails on the first scheduled run.**

`add_server_plugin_api_to_summary.sh` is committed mode `100644`, and the workflow invokes it directly:

```yaml
run: .github/scripts/add_server_plugin_api_to_summary.sh
```

Checked out locally, that gives `permission denied`, exit 126 — and with `set -euo pipefail` the job dies there. `check_links.py` in the same directory is `100755`, so the repo does carry exec bits. Either `git update-index --chmod=+x .github/scripts/add_server_plugin_api_to_summary.sh` or call it as `bash .github/scripts/...`.

**3. shellcheck is red.**

```
.github/scripts/add_server_plugin_api_to_summary.sh:88:11: note: Double quote to prevent globbing and word splitting. [SC2086]
```

Line 88 is `` git diff $summary`` — quote the variable (and drop the stray leading space).

## Correctness

**4. `page-index.md` may not be a file the generator produces.** The script hardcodes it as the section's parent page, links the nav entry at it, and skips it when building the child bullets. But the upstream generator — `include/mysql/generate_plugin_api_docs.sh`, from MDEV-39718 — documents its output as:

```
#   md/api.md - the main plugin API documentation file
#   md/<group>.md - documentation for group <group>
#   md/<page>.md - documentation for each <page>
```

`api.md`, not `page-index.md`. It may well be that moxygen's `--pages` emits `page-index.md` for `include/mysql/index.dox` and the naming is fine — but if it doesn't, the parent nav entry points at a nonexistent page and *nothing catches it*: `test -s "$bullets_file"` still passes, and the intended index page quietly becomes a child bullet instead. Worth a one-line guard either way:

```bash
test -f "$dest_dir/$parent_page"
```

**5. Stefan's second point isn't addressed yet.** He asked that the reviewer of each generated PR get a heads-up when the page set changes. The script rewrites the whole section, so it already knows — diffing the old bullet block against the new one and listing added / removed / renamed pages in the PR body would cover it. The body is static text at the moment.

**6. Nav titles are lifted raw out of `# ` headings.** `title="$(grep '^# ' "$page" | sed ...)"` goes straight into `[%s](%s)`. Doxygen/moxygen headings routinely carry `(`, `)`, `[`, `]` and angle brackets — the page being deleted here is full of `st_mysql_xid`-style names and signatures. Brackets break the link outright, and an unescaped `<...>` in a title renders as nothing in GitBook with no linter complaining. Worth escaping or stripping before writing the bullet.

**7. The `trap` cleanup lines are commented out** (lines 22 and 43), so `mktemp` files leak on the CI path. Worse on the local path: `bullets.tmp` and `tmp_summary.tmp` are written *inside* `server/reference/plugins/api-plugin/` — the directory `add-paths` commits. Restore the traps and use `mktemp` in both branches.

**8. Minor — the old-section removal is order-dependent.** `if (!found && inserted && $0 ~ /\[Plugin API Documentation\]/)` only strips the existing section when it appears *after* the `Plugin Overview` anchor. That holds today, and I traced a second run against the current `server/SUMMARY.md` (anchor at L2189) — it replaces cleanly, no duplication. But if anyone hand-moves the section elsewhere in SUMMARY, the script silently ends up with two copies rather than failing.

**9. Minor — `ghcr.io/mariadb/mariadb-doc-gen:latest` is unpinned.** The image resolves fine (I pulled the manifest anonymously, 200). For a daily job that opens PRs into the docs tree, a digest pin would make a generated diff attributable to a server change rather than an image change.

**10. Note, not yours alone.** A PR opened with the default `GITHUB_TOKEN` doesn't trigger the repo's `on: pull_request` checks, so these generated PRs would arrive without doc-lint, link-check or fragcheck having run. `sync-topical-skills.yml` has the same shape, so it's a repo-wide question — but it bites harder here, where the content is machine-generated and nobody has eyes on it before the bot commits.

## What checks out

- **Deleting `server/reference/plugins/plugin_api.md` is safe.** It appears in no `SUMMARY.md`, nothing in the tree links to it, and `https://mariadb.com/docs/server/reference/plugins/plugin_api` already returns 404 — it was never published. No redirect needed, which is the usual trap with a 7,000-line removal.
- **`sparse-checkout: include/mysql` with cone mode off is exactly right.** `Doxyfile.generated_docs_plugin_api` sets no `INPUT` and runs `RECURSIVE = YES` from that working directory, so nothing outside `include/mysql` is ever read. The comment on that block is accurate.
- **The "same image + script as the server side" claim holds** — `generate-api-docs.yml` and the `ghcr.io/mariadb/mariadb-doc-gen:latest` move both land in `3e3f56d99e0` (MDEV-39806) on `MariaDB/server` `main`. The two will stay in step.
- `actions/checkout@v7` and `peter-evans/create-pull-request@v8` match every other workflow in the repo.

Fix 1–3 and this is close. 4 is the one I'd most want confirmed before the first scheduled run, since it fails silently rather than loudly.

mariadb-stefan-hinz:
Re-reviewed after the push (`0ba962428`). **Blockers 1–3 are resolved**, and CI is green across the board (actionlint, shellcheck, doc-lint on both runners, codespell, fragcheck, includecheck, link-check, expand-links).

Checked against the branch:

- **1 — one workflow now.** `git diff main...pr1012 --name-status` shows only `sync-server-api-plugin-docs.yml` added; the `plugin-api` one is gone from the tree.
- **2 — fixed by invocation.** The script is still committed `100644`, but the step now calls `bash .github/scripts/add_server_plugin_api_to_summary.sh`, so the exec bit no longer matters. (Only leftover: the script's own usage comment still says `./add_server_plugin_api_to_summary.sh /path`, which won't work as written.)
- **3 — shellcheck clean**, locally and in CI.
- **4 — closed.** Confirmed off-thread that `page-index.md` is a real page: `include/mysql/index.dox` declares `@mainpage Plugin API Documentation`, which is where the parent title comes from too. No guard needed.
- **8 — verified rather than assumed.** I ran the script twice against a copy of the current `server/SUMMARY.md` with a synthetic three-page `api-plugin/` directory. Second run is byte-identical to the first — it replaces the section cleanly, correct 4-space parent / 6-space child indentation, anchored after `Plugin Overview` at L2189. Idempotent.

## Two things I'd still fix before merge

**The PR body names the workflow that was just deleted.** Line 108 of `sync-server-api-plugin-docs.yml`:

```
Automated by `.github/workflows/sync-server-plugin-api-docs.yml`.
```

That file no longer exists. Every generated PR would link readers at a dead path.

**Neither label exists in this repo.** The workflow asks for:

```yaml
labels: |
  api-plugin
  plugins
```

The repo's full label set is `bug`, `documentation`, `duplicate`, `enhancement`, `good`, `help`, `invalid`, `question`, `wontfix`, `Replication`, `release-notes`, `doc-tools`, `agent-skills`, `automated`. The two sibling bot workflows — `extract-function-skills.yml` and `sync-topical-skills.yml` — both use `agent-skills` + `automated`. At minimum these PRs won't be labelled consistently with the other automation; `automated` plus `doc-tools` would match the convention, or create the two labels first.

## Retracting one of my own notes

**6 (raw nav titles) doesn't bite — my mistake in framing it as a risk.** I checked the actual title source rather than reasoning about it: all thirteen `@defgroup` titles in `include/mysql`, plus the `@mainpage` title, are plain text with no `[`, `]`, `(`, `)`, `<` or `>`. And in the 7,367-line `plugin_api.md` being deleted — same moxygen pipeline — zero of the 30+ headings contain those characters either. The script only reads each file's first `# ` line, so there's nothing to escape today. Worth remembering only if a future `@page` title picks up punctuation.

**7 (traps) is milder than I said.** On the CI path `basedir` is `.`, so both temp files are `mktemp`s outside the tree and can never be picked up by `add-paths` — the leak is two files in the runner's `/tmp`, discarded with the runner. On the local path it's real: I ran it and `bullets.tmp` was left sitting in `server/reference/plugins/api-plugin/`. Uncommenting lines 22 and 43 is still the fix, but it isn't a merge blocker.

## One new note

`page-index.md` becomes the nav parent, so it's the page readers land on for this whole section. Its source is `index.dox`, which currently reads:

```
- Concepts
- Get started: Building a plugin
- Plugin types
- Plugin services
- [API reference](api.md#)
```

Four of those five are plain text, not links, so the landing page renders as a list of dead-end labels. Understood that the content is fluid for now — but the fifth one, `api.md#`, has an empty trailing fragment that's worth dropping while you're in there.

Minor: `chmod +x .server-src/include/mysql/generate_plugin_api_docs.sh` at line 56 is now redundant, since the container invokes it as `bash generate_plugin_api_docs.sh`.

**9** (unpinned `:latest` image) and **10** (generated PRs get no `on: pull_request` checks, since they're opened with the default `GITHUB_TOKEN`) both still stand, but neither is yours to fix here — 10 is repo-wide.

Fix the PR-body filename and the labels and this looks good to me.

mariadb-stefan-hinz (approved):
Approving on `acdfd2193`.

All three blockers from my earlier pass are resolved — one workflow instead of two, the SUMMARY script invoked via `bash` so the missing exec bit no longer matters, and shellcheck clean. The generated-PR body now names the surviving workflow, and `labels: automated / documentation` both exist, so these PRs stay out of the nightly untriaged digest.

Verified rather than assumed: I ran the SUMMARY script twice against a copy of the current `server/SUMMARY.md` and the second run is byte-identical to the first — the section is replaced cleanly, correct 4-space parent / 6-space child indentation, anchored after `Plugin Overview`. `page-index.md` is confirmed real (`@mainpage Plugin API Documentation` in `include/mysql/index.dox`), the `api.md#` fragment is fine on our side (fragcheck doesn't treat it as a fragment link; lychee resolves it), and deleting `plugin_api.md` is safe — no SUMMARY entry, no inbound links, already 404 live. All checks green.

Two non-blocking leftovers, for whenever the file is next touched: the `basedir != "."` branch still leaves `bullets.tmp` in `api-plugin/` (the `rm -f` there runs before the file is written, so it pre-cleans rather than cleans up — CI never takes that path), and `ghcr.io/mariadb/mariadb-doc-gen:latest` is unpinned, so a generated diff can't be attributed to a server change versus an image change. The wider point that bot-opened PRs don't trigger `on: pull_request` checks is repo-wide, not this PR's to solve.

PR #835, opened 2026-08-04
DOCS-6233 Document RANGE COLUMNS INTERVAL partitioning

mariadb-stefan-hinz:
Documents the `INTERVAL` clause of `PARTITION BY RANGE COLUMNS`, new in MariaDB 13.1 ([MDEV-15621](https://jira.mariadb.org/browse/MDEV-15621)). With an `INTERVAL` clause the server adds partitions automatically as data is written, so a time-ranged table no longer needs a scheduled `ALTER TABLE ... ADD PARTITION`.

Jira: [DOCS-6233](https://mariadbcorp.atlassian.net/browse/DOCS-6233)

## What's here

**New page** — `server/server-usage/partitioning-tables/partitioning-types/range-columns-interval-partitioning.md`

* Syntax: the MariaDB form plus both Oracle spellings (`NUMTODSINTERVAL`, `NUMTOYMINTERVAL`), and the fact that the optional `AUTO` keyword has no effect and is not echoed by `SHOW CREATE TABLE`.
* Requirements and restrictions: one temporal partitioning column; `TIMESTAMP` only with `INTERVAL`; at least one partition; no `MAXVALUE`; positive, non-sub-second interval; at least one day for a `DATE` column; the accepted `NUMTO*INTERVAL` units; no subqueries; not available for `LIST COLUMNS`; subpartitioning limited to `[LINEAR] KEY`/`HASH`; the 8192-partition cap.
* How partitions are added: the transition point, `pN` gap-filling naming, the statements that trigger creation, triggers and replicated row events, and a warning about the counterintuitive part — partitions track the **clock**, not the value being written, so a future-dated row still fails while the partitions are created anyway.
* Transactions and replication: no implicit commit (`ROLLBACK` undoes the rows but keeps the partitions), and the implicit `ADD PARTITION` is not binlogged — the replica reconstructs it from the row events.
* Examples, including real `information_schema.partitions` output.

**Counterparts**

* `partitioning-types-overview.md` + `server/SUMMARY.md` — list the new page.
* `range-columns-and-list-columns-partitioning-types.md` — `TIMESTAMP` columns are accepted only in combination with an `INTERVAL` clause, plus a pointer to the new page.
* `create-table.md` — the `partition_options` grammar listed neither `RANGE COLUMNS` nor `LIST COLUMNS`, so adding the `INTERVAL` form meant adding both `COLUMNS` variants; a matching bullet was added to the partition-methods list.

[...]

PR #1015, opened 2026-09-03
DOCS-6401 Fix three defects in the alias-expansion workflow

mariadb-stefan-hinz:
Fixes the three defects recorded on [DOCS-6401](https://mariadbcorp.atlassian.net/browse/DOCS-6401) in `.github/workflows/expand-gitbook-aliases.yml`.

## 1. The `sed` ran over the whole repository

Any alias that reached the base branch unexpanded — a direct-to-`main` commit is never seen by this workflow — was rewritten by whichever PR opened next, whatever that PR was about. The author then owned a collateral edit to someone else's file. Observed twice on the same file and the same sentence: #835 (a partitioning PR) and #834 (a MaxScale scripts PR that changes no `.md` files at all). On #834 the collateral edit then rotted into a merge conflict on a file the author never touched.

A new step computes the PR's changed Markdown files once, and **both** the expansion step and the `Fail on an unexpanded alias` check read that one list. Previously two `find` invocations had to be kept in sync by comment; now the two file sets are identical by construction.

Scoping the check matters as much as scoping the expansion: with a whole-tree check and a scoped expansion, one unknown alias sitting on `main` in a file nobody is touching would fail **every** PR, with nothing the author could do from their own branch.

The exclusions move from `find` predicates to git pathspecs, so one command produces the list — no shell globbing, no filter pipeline, no dependence on a particular `grep`. The set is unchanged: `dev-docs/` and `.claude/` wholesale, the top-level and `pdf/` READMEs, any `CONTRIBUTING.md`, and `about-links.md`. Every other `README.md` still expands, per DOCS-6481.

## 2. It pushed a merge commit onto the PR branch

`actions/checkout` defaults to `refs/pull/<n>/merge` on a `pull_request` event, so `git push origin HEAD:$BRANCH` put that ephemeral merge commit on the contributor's branch — which this repo, merging by rebase and rejecting `--squash`, cannot take cleanly. It was also unrecoverable by the author: a clean force-push fires `synchronize`, the workflow re-runs, and the merge commit comes back.

Now checks out `${{ github.event.pull_request.head.sha }}` explicitly and pushes `HEAD:refs/heads/$HEAD_REF`.

## 3. The branch name was interpolated into an inline script

```
expand-gitbook-aliases.yml:65:240: "github.event.pull_request.head.ref" is potentially
untrusted. avoid using it directly in inline scripts. instead, pass it through an
environment variable.
```

A branch name is attacker-controlled on a public repository, and this job holds `contents: write` on PR branches. Bound to an `env:` key and quoted in the script — the house rule `link-check-pr.yml` already follows (DOCS-6361).

## Verification

- **`actionlint` clean on this file.** Repo-wide findings go 7 → 6; the remaining six are pre-existing `SC2086` info notes in `codespell.yml` / `generate-pdfs.yml` / `link-check-pr.yml` plus the known `SC2088` false positive (`~/changes/` there is a GitBook URL path, not a home directory). The new shell adds none.
- **The shell logic is tested end to end** in a scratch repo: a PR touching one Markdown file whose path contains a space, a file the PR deletes, and a `dev-docs/` file. In-scope list came out as exactly the one changed file (deletion dropped by `--diff-filter=d`, `dev-docs/` excluded); all four link-target forms expanded (`](`, `]:`, `url="`, `href="`); the leftover check then reported exactly the one unknown alias and nothing else. NUL-delimited throughout, so the space in the path never split.
- **`:(exclude,glob)**/CONTRIBUTING.md` matches at the repo root too** — checked with `git ls-files`, since that is the case a `**/` prefix is easy to get wrong.
- **Swept `main` for the backlog** the ticket asks about, since a scoped expansion no longer cleans up after direct-to-`main` commits: zero unexpanded aliases in link-target position. The only four files matching the pattern are the intentionally excluded ones that document the mechanism (`CONTRIBUTING.md`, `dev-docs/gitbook-syntax.md`, `dev-docs/link-aliases.md`, `.claude/skills/gitbook-format/SKILL.md`). So nothing is left behind by the narrowing.
- **This PR exercises the new version of itself.** Its only changed Markdown file is `dev-docs/link-aliases.md`, which is excluded, so the expected result is an empty file list, the "nothing to expand" path, and no bot commit.

## Side effect worth knowing

The force-push safety note on DOCS-6545 (PR #834's pending rebase) gets simpler once this merges. Both halves of the hazard are gone rather than merely improbable: there is no merge commit to re-add, and #834 changes no `.md` files, so the file list is empty and the expansion cannot fire at all — no longer a conclusion that depends on `main` happening to be free of expandable aliases at the moment of the push.

## Not changed

A PR from a fork still fails at the push step, because `origin` is the base repository. That predates this change and is unaffected by it.

## Drive-by

`dev-docs/link-aliases.md` documents the new scope, including the consequence for authors: an aliased link committed directly to `main` is now never expanded and never reported, and GitBook publishes it verbatim as a `github.com/...` URL that 404s. Touching the file in any PR fixes it.

[DOCS-6401]: https://mariadbcorp.atlassian.net/browse/DOCS-6401?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/.github/workflows/expand-gitbook-aliases.yml
+++ b/.github/workflows/expand-gitbook-aliases.yml
@@ -11,15 +11,77 @@ jobs:
     runs-on: ubuntu-latest
 
     steps:
-      - name: Checkout PR commit
+      - name: Checkout the PR head commit
         uses: actions/checkout@v7
         with:
+          # The head commit, explicitly. The default for a pull_request event
+          # is refs/pull/<n>/merge, and the push step below would then land
+          # that ephemeral merge commit on the contributor's branch -- which
+          # this repository cannot take, because it merges by rebase and
+          # rejects --squash. Worse, it was unrecoverable by the author: a
+          # clean force-push fires "synchronize", the workflow re-runs, and the
+          # merge commit comes straight back (DOCS-6401).
+          ref: ${{ github.event.pull_request.head.sha }}
           fetch-depth: 0
           persist-credentials: true
+
+      - name: List the PR's changed Markdown files
+        env:
+          BASE_REF: ${{ github.event.pull_request.base.ref }}
+        run: |
+          set -eu
+
+          # The expansion below used to walk the whole tree. Any alias that
+          # reached the base branch unexpanded -- a direct-to-main commit is
+          # never seen by this workflow -- was therefore rewritten by the next
+          # PR to open, whatever that PR was about, leaving its author with a
+          # collateral edit to someone else's file in their diff (DOCS-6401,
+          # observed on #835 and #834). So compute the PR's own changed
+          # Markdown files once here, and let both the expansion step and the
+          # unexpanded-alias check consume that one list: the two file sets are
+          # now identical by construction rather than by keeping two find
+          # invocations in sync.
+          #
+          # The exclusions mirror the former find predicates: dev-docs/ and
+          # .claude/ wholesale, the top-level and pdf/ READMEs, any
+          # CONTRIBUTING.md, and about-links.md. Those files hold aliases as
+          # documentation examples. Every other README.md is fair game, per
+          # DOCS-6481. They are git pathspecs rather than a filter pipeline so
+          # that one command produces the list, with no shell globbing and no
+          # dependence on a particular grep.
+          CHANGED_MD="$RUNNER_TEMP/changed-md.nul"
+
+          git fetch --no-tags --quiet origin \
+            "+refs/heads/$BASE_REF:refs/remotes/origin/$BASE_REF"
+
+          # NUL-delimited, so a path containing a space cannot split.
+          # --diff-filter=d drops files the PR deleted; they cannot be sed-ed.
+          git diff --name-only --diff-filter=d -z \
+            "refs/remotes/origin/$BASE_REF...HEAD" -- \
+            '*.md' \
+            ':(exclude)dev-docs/' \
+            ':(exclude).claude/' \
+            ':(exclude)README.md' \
+            ':(exclude)pdf/README.md' \
+            ':(exclude,glob)**/CONTRIBUTING.md' \
+            ':(exclude,glob)**/general-resources/about/readme/about-links.md' \
+            > "$CHANGED_MD"
+
+          echo "CHANGED_MD=$CHANGED_MD" >> "$GITHUB_ENV"
+
+          echo "Markdown files in scope for alias expansion:"
+          tr '\0' '\n' < "$CHANGED_MD"
+
       - name: Expand GitBook aliases (Stable Version)
         run: |
-          set -e
-          echo "Expanding GitBook aliases in link targets, excluding protected files..."
+          set -eu
+
+          if [ ! -s "$CHANGED_MD" ]; then
+            echo "No Markdown files changed in this PR; nothing to expand."
+            exit 0
+          fi
+
+          echo "Expanding GitBook aliases in link targets..."
 
           # An alias is only rewritten where it is a link target: directly after
           # a Markdown link's "](", after a reference-style link definition's
@@ -36,14 +98,7 @@ jobs:
           #
           # Keep the prefix group in sync with the "Fail on an unexpanded alias"
           # step below, which reports what this step could not expand.
-          find . -type f -name "*.md" \
-            ! -path "./dev-docs/*" \
-            ! -path "./.claude/*" \
-            ! -path "./README.md" \
-            ! -path "./pdf/README.md" \
-            ! -name "CONTRIBUTING.md" \
-            ! -path "*/general-resources/about/readme/about-links.md" \
-            -exec sed -E -i \
+          xargs -0 -r sed -E -i \
             -e 's#(^[[:space:]]*\[[^]]*\]:[[:space:]]*|\]\(|url="|href=")[{]server[}]#\1https://app.gitbook.com/o/diTpXxF5WsbHqTReoBsS/s/SsmexDFPv2xG2OTyO5yV#g' \
             -e 's#(^[[:space:]]*\[[^]]*\]:[[:space:]]*|\]\(|url="|href=")[{]galera[}]#\1https://app.gitbook.com/o/diTpXxF5WsbHqTReoBsS/s/3VYeeVGUV4AMqrA3zwy7#g' \
             -e 's#(^[[:space:]]*\[[^]]*\]:[[:space:]]*|\]\(|url="|href=")[{]maxscale[}]#\1https://app.gitbook.com/o/diTpXxF5WsbHqTReoBsS/s/0pSbu5DcMSW4KwAkUcmX#g' \
@@ -57,12 +112,21 @@ jobs:
             -e 's#(^[[:space:]]*\[[^]]*\]:[[:space:]]*|\]\(|url="|href=")[{]mariadb-cloud[}]#\1https://app.gitbook.com/o/diTpXxF5WsbHqTReoBsS/s/vPz15Lz0Iw3P3yKR3Prd#g' \
             -e 's#(^[[:space:]]*\[[^]]*\]:[[:space:]]*|\]\(|url="|href=")[{]release-notes[}]#\1https://app.gitbook.com/o/diTpXxF5WsbHqTReoBsS/s/aEnK0ZXmUbJzqQrTjFyb#g' \
             -e 's#(^[[:space:]]*\[[^]]*\]:[[:space:]]*|\]\(|url="|href=")[{]general-resources[}]#\1https://app.gitbook.com/o/diTpXxF5WsbHqTReoBsS/s/WCInJQ9cmGjq1lsTG91E/about/readme#g' \
-            {} +
+            < "$CHANGED_MD"
 
           echo "Git status after expansion:"
           git status --short
+
       - name: Commit and push changes back to PR branch
+        env:
+          # Not interpolated into the script body: a branch name is
+          # attacker-controlled on a public repository, and this job holds
+          # write access to PR branches (DOCS-6401, and the house rule
+          # link-check-pr.yml already follows, DOCS-6361).
+          HEAD_REF: ${{ github.event.pull_request.head.ref }}
         run: |
+          set -eu
+
           if git diff --quiet; then
             echo "No alias replacements needed."
             exit 0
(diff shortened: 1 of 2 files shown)


mariadb-stefan-hinz:
**Both paths verified in CI, not just the quiet one.**

This PR's own run only exercised the empty-list branch — its single changed Markdown file, `dev-docs/link-aliases.md`, is excluded, so the log reads "No Markdown files changed in this PR; nothing to expand", no bot commit, check skipped. That proves the guard, not the fix. A check that never fires can't prove itself, so I ran the firing path on a throwaway branch (#1016, now closed and deleted) carrying one scratch page with an aliased link.

Result, from run `33779787965`:

```
Markdown files in scope for alias expansion:
general-resources/zzz-docs-6401-scratch.md
 M general-resources/zzz-docs-6401-scratch.md
Pushing to PR branch: DOCS-6401-scratch-verify
No unexpanded aliases.
```

The alias expanded correctly, and the branch's first-parent chain came back **linear**:

```
9cd0b7c83  github-actions — docs: expand GitBook aliases
4cd1cb48c  Stefan Hinz    — DOCS-6401 scratch: verify …
adb1e714e  Stefan Hinz    — DOCS-6401 Fix three defects …
```

`git log --merges` over the branch: **0**. That is precisely where the old version inserted a `Merge <head> into <base>` merge-ref artifact, so defect 2 is fixed in fact and not only in principle.

One incidental confirmation. The scratch page also carried `A reference-style one: [Securing]: {maxscale}/…` mid-line, which was **not** expanded — correctly, since a reference-style definition must begin its line, and the regex requires `^[[:space:]]*`. The unexpanded-alias check didn't flag it either, because it applies the same anchor. Worth noting only because the two patterns staying consistent is the property this PR now guarantees structurally rather than by comment.

dbart (changes requested):
All three fixes are the right ones, and the verification section holds up — I re-checked the claims I could rather than taking them on trust, and every one of them was accurate (details at the bottom). One change requested, in the step this PR rewrites.

## Requested change: the failure message loses the filename on a single-file PR

In `Fail on an unexpanded alias`:

```sh
LEFTOVER=$(xargs -0 -r grep -nE '...' < "$CHANGED_MD" || true)
```

`grep -n` prints `file:line:text` when it gets **several** files, but only `line:text` when it gets **one**. The old code ran `find` across the whole tree, so it always passed hundreds of paths and the filename was always present. Scoping to the PR's own changed files makes the one-file invocation ordinary — and a single changed Markdown file is the most common PR shape in this repo.

Running your exact invocation against a file containing an unknown alias:

```
# one file in $CHANGED_MD
::error::Unknown GitBook alias left unexpanded in a link target
1:ok [b]({bogusalias}/y)

# two files in $CHANGED_MD
docs/deep/nested/with space.md:1:ok [b]({bogusalias}/y)
```

The step still fails correctly, so nothing ships broken — it just stops saying *where*, in what the comment three lines above rightly calls "the only place an author sees it." On a PR touching one file the author can infer it; the cost is real on a rename or a split where the alias could be in any of several files, and it's a silent downgrade from the behaviour this step had before.

`-H` fixes it:

```sh
LEFTOVER=$(xargs -0 -r grep -nHE \
  '(^[[:space:]]*\[[^]]*\]:[[:space:]]*|\]\(|url="|href=")[{][A-Za-z0-9_-]+[}]' \
  < "$CHANGED_MD" || true)
```

That also covers the corner where `xargs` batch-splitting happens to leave one file in the final batch. Verified with the same harness.

## Not blocking: the whole-tree guard is worth replacing

Narrowing the expansion is clearly right, and the PR is straight about the consequence — an alias reaching `main` outside a PR is now never expanded and never reported. The sweep showing a clean `main` today is good evidence nothing is stranded right now, but it's a point-in-time check, and the mitigation ("touching the file in any PR expands it") needs somebody to notice first.

This repo has already made the argument, in `nightly-fragcheck.yml`'s own header (DOCS-6524):

> This repo has write paths that never produce a pull request, and a `pull_request` trigger is structurally blind to all of them: GitBook-UI edits sync straight back to Git as `GITBOOK-*` commits […] So the gate covers PRs and this covers the tree.

That reasoning transfers to aliases exactly. GitBook-UI editing is a first-class write path here, and a writer inserting `{server}/foo` in the web app produces precisely the plausible-looking `github.com/...` URL that 404s — the thing this check exists to catch, with lychee excluding braced URLs so nothing downstream sees it either. A nightly whole-tree alias digest mirroring the fragcheck pattern (same invocation, different base) would close it.

Filed as [DOCS-6588](https://mariadbcorp.atlassian.net/browse/DOCS-6588) and linked to DOCS-6401, so it stays off this PR's critical path — nothing for you to do about it here. It carries the sweep numbers, the exclusion set to reuse, and the "rolling base, not an absolute inventory" constraint from DOCS-6524, so whoever picks it up doesn't have to rederive any of it.

## Verified independently

| Claim | Result |
| --- | --- |
| `actionlint` 7 → 6, new shell adds none | Confirmed. v1.7.12: 7 on `main` @ `aafc6c0db`, 6 at `adb1e714e`; the one removed is `expand-gitbook-aliases.yml:65:240`, the exact finding quoted |
| Remaining 6 are pre-existing, incl. an SC2088 false positive | Confirmed — `~/changes/` in `link-check-pr.yml` is a `git grep` pattern, not a shell tilde |
| Pathspecs match the former `find` predicates | Confirmed via `git ls-files` |
| `**/CONTRIBUTING.md` matches at the repo root too | Confirmed — root and nested both excluded |
| `--diff-filter=d` drops deletions | Confirmed (lowercase excludes) |
| `main` has no unexpanded aliases in scope | Confirmed — 0 in-scope across 9,708 tracked `.md` |
| "Only four files match, all intentionally excluded" | Confirmed — exactly 4 files / 7 hits: `CONTRIBUTING.md`, `dev-docs/gitbook-syntax.md`, `dev-docs/link-aliases.md`, `.claude/skills/gitbook-format/SKILL.md` |
| This PR exercises its own new version | Consistent — its only `.md` is excluded, so the empty-list path ran and `expand-links` passed |

Fixes 2 and 3 read as unambiguously correct: `ref: head.sha` plus `git push origin "HEAD:refs/heads/$HEAD_REF"` removes the merge-commit-onto-the-contributor's-branch trap, and binding `HEAD_REF` / `BASE_REF` through `env:` is the right remedy on a job holding `contents: write`. Deriving the file list from `git diff` with pathspecs rather than a third-party action is also a nice reduction in what this write-privileged job depends on, compared with `fragcheck-pr.yml`.

Add `-H` and I'll approve.

mariadb-stefan-hinz:
**`-H` added in `2a068f2f9`** — you were right, and the reproduction matches yours exactly.

Confirmed the behaviour before changing anything, on both grep flavours that matter here, since CI runs GNU grep and this machine's `grep` is ugrep:

```
# GNU grep 3.12 and ugrep 7.8.4, one-file list, -nE
1:ok [b]({bogusalias}/y)

# same list, -nHE
docs/deep/nested/with space.md:1:ok [b]({bogusalias}/y)
```

So the downgrade was real on the runner too, not only locally.

I also checked whether the same shape sits anywhere else in `.github/workflows/`. The only other `-n` search is `git grep -n` in `link-check-pr.yml`, which is unaffected — `git grep` always emits the path, because it searches the tree rather than an argument list. Nothing else to fix.

The flag carries a comment saying why, because a bare `-H` is exactly the kind of thing that reads as redundant later and gets tidied away — it names the one-file-vs-several behaviour, the `xargs` final-batch corner you pointed out, and the two grep versions it was verified on.

All 7 checks green. `expand-links` exercised the empty-list path again, as expected — this PR's only changed file is a workflow, so no Markdown is in scope. The firing path stays proven by the throwaway run on #1016 in the comment above.

Thanks for the independent verification table — and for filing [DOCS-6588](https://mariadbcorp.atlassian.net/browse/DOCS-6588) rather than parking the nightly-digest point on this PR. The `nightly-fragcheck.yml` precedent is the right one: a `pull_request` trigger is structurally blind to GitBook-UI writes, and an alias inserted in the web app produces precisely the 404-ing `github.com/...` URL this check exists to catch, with lychee excluding braced URLs so nothing downstream sees it either.


[DOCS-6588]: https://mariadbcorp.atlassian.net/browse/DOCS-6588?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

dbart (approved):
Approved. `-H` in `2a068f2f9` is exactly the fix, and the comment above it is the part that will keep it alive — a bare `-H` is precisely what gets tidied away later as redundant.

Re-checked at the new head:

- **actionlint: 6**, unchanged from main's baseline, so the fix introduces nothing. The one finding that disappeared between `aafc6c0db` and this branch is still the `head.ref` script-injection one this PR set out to remove.
- **Your sweep claim holds.** I went through every `grep` in all eleven workflows: the only other `-n` search is the `git grep -n` in `link-check-pr.yml`, and `git grep` does emit the path even with a single pathspec — checked rather than assumed. `generate-pdfs.yml` uses `-aoE` off a pipe, so the filename question doesn't arise there.
- 7/7 green, `MERGEABLE` / `CLEAN`. The two new jobs are inherited from `doc-lint-test.yml` landing on `main` in DOCS-6471, not from anything here — `pull_request` evaluates workflows from the merge ref.

The #1016 run is the part worth calling out, because it reaches something review structurally could not. This PR can only ever exercise the empty-list branch — its sole Markdown file is excluded — so the firing path was unprovable from the diff. A linear first-parent chain with `git log --merges` at 0, on exactly the branch shape where the old version inserted the merge-ref artifact, settles defect 2 as fact rather than reasoning. Building a throwaway branch to make a check prove itself is the right instinct; "a check that never fires can't prove itself" is worth keeping as a general rule.

Good catch on the mid-line reference-style definition not expanding, too. That the expansion and the leftover check now share one anchor by construction rather than by a comment asking someone to keep two `find` calls in sync is the quiet structural win in this PR, and it's what makes that consistency a property instead of a convention.

Nightly-digest point is [DOCS-6588](https://mariadbcorp.atlassian.net/browse/DOCS-6588), linked to DOCS-6401 and sitting in the backlog — off this PR's path entirely.

Merge and DOCS-6401 are yours.

PR #1016, opened 2026-09-03
DO NOT MERGE — DOCS-6401 scratch verification

mariadb-stefan-hinz:
Throwaway. Verifies that the reworked expand-gitbook-aliases workflow (#1015) actually expands an alias and pushes the bot commit **without** a merge commit. Will be closed and the branch deleted as soon as the run is read.

### DIFF
--- a/.github/workflows/expand-gitbook-aliases.yml
+++ b/.github/workflows/expand-gitbook-aliases.yml
@@ -11,15 +11,77 @@ jobs:
     runs-on: ubuntu-latest
 
     steps:
-      - name: Checkout PR commit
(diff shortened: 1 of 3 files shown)


mariadb-stefan-hinz:
Verification done, discarding as planned. The reworked workflow behaved exactly as intended on this scratch branch: the in-scope list was the one changed file, the alias expanded, the bot pushed, and the branch's first-parent chain is linear — `9cd0b7c83 github-actions` sitting directly on `4cd1cb48c`, with zero merge commits. Under the old version this is the point at which a `Merge X into Y` merge-ref artifact appeared. See #1015.

PR #1017, opened 2026-09-03
DOCS-6471 Add CI coverage for doc-lint.sh, and fix the bug it found

mariadb-stefan-hinz:
[DOCS-6471](https://mariadbcorp.atlassian.net/browse/DOCS-6471)

Nothing in CI ran `.claude/hooks/doc-lint.sh`, and its only caller is a Claude Code `PreToolUse(Bash)` hook that — as `pre-commit.sh`'s own header says — "only gates commits that Claude Code makes via the Bash tool. Human `git commit`, IDE commits, and GitBook-UI edits bypass it entirely." So the script hosting the **only** checks for dead GitBook includes (DOCS-6372) and silently gutted pages (DOCS-6470, for the DOCS-6442 class) was covered by nothing at all.

DOCS-6409 is what that cost: a `mktemp` template with no `X` characters made the script exit 2 on every GNU-coreutils run for **135 commits**, and because the failing line sat *before* the checks, it silently took the include resolver down with it — so a feature added the same day never ran on Linux even once.

This is **Part 1** of the ticket, the primary ask. Part 2 is deliberately split out (see below).

## What's here

**`.claude/hooks/doc-lint-test.sh`** — 33 fixture cases, runnable locally, building their fixtures in a throwaway git repo under `TMPDIR`. It asserts both the exit code and the message for the ticket's whole table plus rather more: the include resolver (resolvable, dead, cross-space, `https`, the `dev-docs/`+`.claude/` exemption, multiple failures in one run), the shrink guard (gutted, `DOC_LINT_ALLOW_SHRINK` by path / by list / `=all`, absent from the base rev, each of the three thresholds isolated one at a time, an unresolvable `DOC_LINT_BASE`), the repo-root guard, the argument filter, and every "tool not installed" SKIP branch.

Two rules come straight from how DOCS-6409 hid, both of which the ticket calls for:

- **Nothing is piped.** Per DOCS-6409 impact note 4, `doc-lint.sh … | tail` returns the *pipe's* status, so a hard failure reads as a clean 0. Every case redirects to files and reads `$?` directly — and so does the workflow step that runs the suite.
- **The SKIP branches are asserted**, and so are the *tool-present* branches. A script that dies in its own plumbing before the first `command -v` passes every SKIP assertion and still checks nothing — that is exactly what DOCS-6409 did. Three cases prove codespell is actually reached (a misspelling fails, `-I .codespellignore` is honoured, `SUMMARY.md` is excluded as `codespell.yml` does) and three prove lychee is.

**`.github/workflows/doc-lint-test.yml`** — two jobs: shellcheck over every tracked shell script, and the suite on `ubuntu-latest` **and** `macos-latest`.

## Three design calls worth a look

**It runs on every PR, not only when the hooks change.** DOCS-6409 was not caused by an edit to the script — it was caused by the script meeting different coreutils. A `paths:` filter on `.claude/hooks/**` would have been green throughout those 135 commits. The suite reads nothing from the checkout (it builds its own repo in `TMPDIR`), so unlike a content check it cannot hand an unrelated PR's author a finding they are unable to clear from their own branch — the failure mode DOCS-6401 documents.

**Both platforms.** `macos-latest` still ships bash 3.2, which is the portability floor `doc-lint.sh` claims in its header. That claim was already broken — see below.

**shellcheck and lychee are pinned by version + sha256** rather than taken from the runner image, so a runner bump cannot add a new check and turn unrelated PRs red on files their author never touched. The shellcheck pin matches what the ticket measured clean (0.11.0). Note the lychee pin is independent of `link-check-pr.yml`'s: that workflow gates the link check at whatever its action ships, this one only feeds the fixture suite, so the two need not move together.

The shellcheck target list is **discovered** (`git ls-files '*.sh'`), not an allowlist, so a script added anywhere is gated the day it lands.

## The bug the suite found on its first run

Expanding `"${arr[@]}"` or `"${arr[*]}"` on an **empty** array is an "unbound variable" error under `set -u` before bash 4.4 — i.e. on the bash 3.2 that stock macOS still ships. In `norm_path`, a `..` that pops the last remaining segment emptied the array and the recompaction died, so the function returned the **empty string**:

```
$ /bin/bash -c 'set -u; ...; norm_path "server/../maxscale/included.md"'
line 185: out[@]: unbound variable
[]
```

Two consequences. Any include climbing out of its own directory was misreported as *unresolvable*. And since escaping the top-level directory is precisely what a cross-space include has to do, **the cross-space branch was unreachable on macOS entirely** — the check the ticket lists as its own fixture row had never once fired there.

The mirror image of DOCS-6409: same class of bug, opposite platform, invisible to a single-runner gate. Which is the argument for the matrix, made by the thing itself.

No page in the repo hits it today — I checked every `{% include %}` in the tree for one with as many leading `..` segments as its page has directory levels, and there are none — so this clears a latent defect rather than a live failure.

## Test plan

Ran locally on macOS (bash 3.2, shellcheck 0.11.0, codespell, lychee 0.24.2):

- `.claude/hooks/doc-lint-test.sh` → **33 passed, 0 failed, 0 skipped**
- `shellcheck` over the discovered list → clean; `actionlint` on the new workflow → clean
- `codespell` (CI-exact invocation) and `doc-lint.sh` over the two changed Markdown files → clean
- `fragcheck.py new origin/main` → no new dead anchors, no new stolen ids

**And the gate was proved to fail, not just to pass.** A check that has never gone red proves nothing (DOCS-6481, DOCS-6401), so I mutated `doc-lint.sh` twelve ways and confirmed each is caught:

| Mutation | Result |
|---|---|
| `mktemp` fails before the checks — the DOCS-6409 shape | 4 passed, **28 failed** |
| include resolver never tallies a failure | **4 failed** |
| cross-space branch removed | **2 failed** |
| the `norm_path` fix reverted | **2 failed** |
| shrink guard never fails | **2 failed** |
| `DOC_LINT_ALLOW_SHRINK` ignored | **3 failed** |
| repo-root guard removed | **1 failed** |
| a codespell SKIP notice turned into a hard failure | **18 failed** |
| lychee "No links were found" carve-out removed | **1 failed** |
| `SUMMARY.md` no longer excluded from codespell | **1 failed** |
| `dev-docs/` + `.claude/` include exemption removed | **1 failed** |
| the "file must exist" argument filter removed | **1 failed** |
| *(control: unmutated)* | 33 passed, 0 failed |

Two of those needed the suite tightened before they were caught, and both are worth knowing about:

- **The lychee carve-out is a compatibility shim.** `failIfEmpty: false` is a *lychee-action* option — the action runs its own emptiness check, while the lychee **binary** has exited 0 on a link-free file since ~0.16 (verified on 0.24.2). So no installed lychee can reach that branch. A stub `lychee` on `PATH` pins it, which keeps the shim from rotting into dead code that a future cleanup removes by accident rather than on purpose.
- **The argument filter's `[ -f "$f" ]` is load-bearing only when a checker is present.** Without it, a path deleted or renamed since it was staged reaches lychee, which exits 1 with "Input not found as file and not a valid URL" — reported to the author as a broken *link*, on a page that no longer exists. With both checkers absent the unfiltered path is inert, so only a tool-present case catches it.

## Deliberately not in this PR

**Part 2** — running the include resolver and the shrink guard as a PR gate over repo content. The ticket calls it optional and larger, and it is: `DOC_LINT_ALLOW_SHRINK` is an environment variable no PR author can set from a branch, so it needs an in-band acknowledgment path (a PR label, a commit-message trailer, or a checked-in allowlist), and choosing among those is a real decision rather than an implementation detail. Filing as a follow-up.

**`release-notes/maxscale/script/*.sh`** is excluded from the shellcheck gate, with the reason on the record in the workflow rather than as a bare exclusion. It carries 12 pre-existing findings, all info/warning-level, and one is more than style: `upgrade_version`'s comment says it "needs to be updated whenever a new major release is out", but the heredoc hardcodes the upgrading link and never reads the variable, so the instruction is misleading to whoever next cuts a release. Fixing it changes generated release-note output, which is the owning team's call, not something to slip into an infrastructure PR — reported on the ticket instead. Drop the exclusion once it is resolved.

**`actionlint` in CI** — adjacent and worth having (it embeds shellcheck for `run:` blocks and flags the untrusted-input class DOCS-6401 fixed by hand), but it needs its own decision about the 6 pre-existing findings on `main` and the known `SC2088` false positive in `link-check-pr.yml`. Separate ticket.

[DOCS-6471]: https://mariadbcorp.atlassian.net/browse/DOCS-6471?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/.claude/README.md
+++ b/.claude/README.md
@@ -9,6 +9,7 @@ Code. It contains:
 | `settings.local.json` | **Personal** overrides — gitignored, never committed |
 | `hooks/doc-lint.sh` | Canonical codespell + lychee linter (single source of truth, mirrors CI) |
 | `hooks/pre-commit.sh` | PreToolUse hook: gates Claude-made `git commit`s by calling `doc-lint.sh` |
+| `hooks/doc-lint-test.sh` | Regression suite for `doc-lint.sh` — fixtures in a throwaway repo; run it after editing the linter |
 | `skills/` | Shared skills (e.g. `docs-check`) |
 | `commands/` | Shared slash commands (e.g. `/precommit`) |
 
@@ -98,6 +99,31 @@ team-wide guarantee. For coverage of *all* commits, install a real Git hook
 A file changed earlier in your branch but not currently staged can pass the hook yet still be
 linted by CI. Before opening a PR, run the full check over your whole branch diff — see
 `dev-docs/cookbook-pre-pr.md`.
+
+## Who tests the linter
+
+`hooks/doc-lint.sh` had no test of its own until DOCS-6471, and it cost a real outage: DOCS-6409
+was a `mktemp` template with no `X` characters that made the script **exit 2 on every
+GNU-coreutils run** — for 135 commits. Because the failing line sat *before* the checks, it
+silently took the GitBook include resolver down with it, so a feature shipped the same day never
+ran on Linux at all. On BSD/macOS `mktemp` the identical line worked fine, so the only machine
+that could have noticed was a Linux one, and nothing on Linux ever ran the script.
+
+`hooks/doc-lint-test.sh` is the gate for that class. It builds fixtures in a throwaway git repo
+under `TMPDIR` — it never reads this checkout's content — and asserts exit codes and messages for
+the include resolver, the gutted-page guard, the env-var knobs, the repo-root guard, and every
+"tool not installed" SKIP branch. Run it after any change to `doc-lint.sh`:
+
+```bash
+.claude/hooks/doc-lint-test.sh              # --keep to inspect the sandbox, --verbose for output
+```
+
+`.github/workflows/doc-lint-test.yml` runs it on **every** PR, on both `ubuntu-latest` and
+`macos-latest`, and shellchecks every tracked shell script alongside it. Both platforms on
+purpose: the suite's first run found the mirror image of DOCS-6409 — an empty-array expansion
+that is an error under `set -u` before bash 4.4, so on the bash 3.2 that stock macOS still ships,
+every `{% include %}` climbing out of its own directory was misreported as unresolvable. Same
+class of bug, opposite platform. A single-runner gate sees one and not the other.
 
 ## Single source of truth
 
--- a/.claude/hooks/doc-lint-test.sh
+++ b/.claude/hooks/doc-lint-test.sh
@@ -1,0 +1,551 @@
+#!/usr/bin/env bash
+#
+# doc-lint-test.sh — regression suite for doc-lint.sh.
+#
+# Usage:   .claude/hooks/doc-lint-test.sh [--keep] [--verbose]
+#          --keep     leave the sandbox behind and print its path
+#          --verbose  print the captured stdout/stderr of every case
+# Exit:    0 = every case passed, 1 = at least one case failed,
+#          2 = the suite could not set itself up (missing doc-lint.sh, no git, ...)
+#
+# WHY THIS EXISTS
+#   Nothing used to run doc-lint.sh outside the Claude Code pre-commit hook, and that hook
+#   "only gates commits that Claude Code makes via the Bash tool" (pre-commit.sh header) — so
+#   human, IDE and GitBook-UI commits never invoked it. DOCS-6409 is what that costs: a mktemp
+#   template with no X characters made the script exit 2 on every GNU-coreutils run from
+#   f2b5e332c (2026-08-05) until 2026-08-18 — 135 commits — and because the failing line sat
+#   BEFORE the checks, it silently took the include resolver down with it. On macOS/BSD mktemp
+#   the same line worked fine, so the only machine that could have noticed was a Linux one, and
+#   no Linux ever ran the script.
+#
+#   That is the failure class this suite is aimed at: the script's own plumbing, on a platform
+#   the author is not using. It is wired into CI by .github/workflows/doc-lint-test.yml, which
+#   runs it on ubuntu-latest on every PR — deliberately NOT only when the hooks change, because
+#   DOCS-6409 was not caused by an edit to the script. It was caused by the script meeting
+#   different coreutils.
+#
+# WHAT IT COVERS AND WHAT IT DOES NOT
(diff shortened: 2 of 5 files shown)


mariadb-stefan-hinz:
### CI verification — the new checks did real work, not vacuous work

The three new checks are green, but green is not the interesting part: a gate that passes because it checked nothing looks identical from the outside. So, from the run logs:

**`doc-lint-test (ubuntu-latest)`** — bash **5.2.21**, `33 passed, 0 failed, 0 skipped`. The zero skips are load-bearing. The suite reports a case group as SKIP when its optional tool is absent, and `27 passed, 0 failed, 2 skipped` would look green at a glance while silently having tested six fewer things. CI sets `DOC_LINT_TEST_REQUIRE_ALL=1` on this runner, which turns any skip into a failure — so `0 skipped` is proof that `pipx install codespell` (2.4.3) and the pinned lychee (0.24.2) both landed on `PATH` and that every tool-present case ran. Same shape of assertion as the empty-list trap in DOCS-6401: the guard passing tells you nothing about the path that fires.

**`doc-lint-test (macos-latest)`** — bash **3.2.57(1)-release**, `27 passed, 0 failed, 2 skipped`, the two skips being the codespell and lychee groups that are deliberately not installed there. This is the runner that matters for this PR: `ok a cross-space include fails with "cross-space include"` appears in that log, and before the `norm_path` fix in this PR it could not have. That one line is the whole argument for the matrix.

**`shellcheck`** — pinned 0.11.0 (verified by sha256), and the log prints the discovered file list so the gate's scope is visible rather than implied:

```
Checking:
  .claude/hooks/doc-lint-test.sh
  .claude/hooks/doc-lint.sh
  .claude/hooks/pre-commit.sh
  pdf/fetch-deps.sh
```

Four files, clean — the ticket's two hooks plus the new suite plus `pdf/fetch-deps.sh`, which was already clean and came along for free. The two MaxScale release-note scripts are the documented exclusion.

The other four checks (codespell-check, expand-links, fragcheck, link-check) also pass.

PR #1030, opened 2026-09-06
add tidesql storage engine documentation

guycipher:
new docs section for tidesql, the storage engine plugin for mariadb built on the tidesdb library.

- landing page plus 22 pages under server-usage/storage-engines/tidesql
- getting started covers loading the plugin, the install.sh builder, the initial-exec tls allocator gotcha, and packaging via mariadb foundry without the tidesdb installer
- versions and compatibility carries the tested mariadb matrix and the tidesql-major to tidesdb-library mapping (5.x links tidesdb v10)
- concepts for the data model, mvcc transactions and isolation, and the durability sync modes
- reference for table options, all 42 system and 54 status variables, encryption, ttl, full-text, vector and spatial indexes, generated columns and json, and foreign keys
- administration for backup and checkpoint, maintenance, monitoring, online ddl, partitioning, and replication and high availability
- wired into server/summary.md nav and the storage-engines index


guycipher:
Any questions,do let me know, by the way!  I've tested locally when working on converting from repo /doc to gitbook.  
https://github.com/tidesdb/tidesql/tree/5.0.0/doc

Thank you kindly/

mariadb-stefan-hinz:
Thanks for this, Alex — reviewed and merging.

I fact-checked the reference material against `tidesdb/tidesql` and `tidesdb/tidesdb` rather than the upstream manual, and it held up unusually well:

- **42/42 system variables** and **54/54 status variables** match the registered arrays in `ha_tidesdb.cc` and `ha_tidesdb_status.cc` exactly — no invented names, none missing, and every default, scope, and dynamic flag correct.
- **15/15 table options** match `tidesdb_table_option_list`, and `ENCRYPTION_KEY_ID` (default 1, range 1–255) is right.
- All 36 `ENGINE=` examples correctly use `TIDESDB`, and the `SHOW ENGINES` comment string matches `PLUGIN_DESCRIPTION` verbatim. `XA: YES` and `Savepoints: YES` are both borne out by the handlerton registration.
- The `plugin-maturity=beta` explanation is right, including that a stable server defaults the threshold to `gamma` — confirmed on a live 12.3.3 server.

Two changes in the commit on top:

1. `tidesql-versions-and-compatibility.md` said 5.0.0 "links TidesDB v10.0.0". `VERSIONING.md` says v10.**x.x**, and v10.0.1 has already shipped, so I generalized it. (The 5.x→v10.x / 4.x→v9.x table itself checks out against the tag history.)
2. The landing page said moving a table to TidesDB is "a change to the `ENGINE` clause and nothing more" — since the engine rejects two foreign-key shapes outright, I qualified it and linked Limitations.

I also rebased onto current `main`: the branch was 52 commits behind, and `main` has since picked up heading-anchor fixes to the generated `api-plugin` pages, which made the anchor check fail on files your PR never touched.

**One thing for you, not a blocker:** in the MariaDB compatibility table, the `12.2.2` row gives a minimum TideSQL version of `1.0.0` while `11.4.10` needs `3.4.0` and `12.3.1` needs `4.2.6`. That is transcribed faithfully from `doc/06-appendix/compatibility.md`, so I have left it as-is, but it looks like an upstream typo. Worth a look upstream — happy to patch the docs page once you confirm the intended value.

PR #1038, opened 2026-09-08
DOCS-6602 Added triage-epic skill

IgGusev:


mariadb-stefan-hinz:
Reviewed at Stefan's request. **Overall: yes, this makes sense** — the design is sound and unusually well-evidenced. I spot-checked the empirical claims against live Jira and a local `MariaDB/server` clone; almost all of them reproduce exactly. One macOS-only defect below is a blocker; the rest are small.

## What I verified (and it holds)

- **DOCS-6084 breakdown.** `parent = DOCS-6084` returns exactly 28 children: 14 Closed, 11 On Hold, 1 Review (DOCS-6258), 1 TODO (DOCS-6275), 1 DUPLICATED (DOCS-6217) — matching §8's example verbatim. `issuelinks` is indeed empty on the epic, so the `parent` JQL is the right call.
- **Every row of the §6 example table is live-accurate:** MDEV-39518 Closed/Fixed/`13.1.1`; MDEV-40030 In Testing/`13.1`; MDEV-38983 In Review/`13.1`; MDEV-30645 Stalled/`13.2`; MDEV-25292 In Testing/`13.1`.
- **The MDEV-38975 case that justifies "merged ≠ shipped"** is real: merged to `preview-13.1-preview` 2026-06-13, and Elena Stepanova commented 2026-09-04 *"I'm afraid this feature won't make it to 13.1 release."* That single example carries the whole design argument, and it carries it.
- **The MDEV-25292 ordering trap** reproduces precisely: `git log main --grep` lists `1f85eeeb53a` (author date 2022-08-31) *above* `2bd41fc5bf7 Revert MDEV-25292 …` (2022-10-27). "Is the newest commit a revert?" answers *no* here. Good catch, and worth the paragraph it gets.
- **The MDEV-35915 false-positive reverts** are real — two commits titled `Revert "MDEV-37686 rpl.create_or_replace_mix2 fails in MDEV-35915 branch"`.
- The `.issues.nodes[]` spill shape in §2 is correct; the `jira.mariadb.org` preflight returns 200 anonymously; `MariaDB/server` has ~3,995 branches, matching the cost warning.

Separating **"code exists"** from **"it shipped"**, and refusing to promote on merge evidence, is the right call — and "a run that promotes nothing is a good run" is the sentence that will keep this skill honest.

## Blocker: the `git log` snippets silently match nothing on macOS

Every commit query uses `-E` with `\b`:

```bash
git -C "$p" log "$ref" --oneline -E --grep="\bMDEV-40030\b"
```

`\b` is a GNU-regex extension. Under `-E` (POSIX ERE) on macOS/BSD libc it matches **zero** commits — no error, no warning. On this machine (git 2.55.0, macOS):

```
-E with \b : 0 commits
-E plain   : 116 commits
-P with \b : 116 commits
```

The revert query is the dangerous one. `git log --all -E --grep="^Revert.*\bMDEV-25292\b"` returns **nothing** here, so §4b's headline guardrail — *"any revert disqualifies"* — silently never fires, and MDEV-25292's real revert (`2bd41fc5bf7`) sails through as clean. The commit-evidence queries fail the same way, which quietly demotes every CANDIDATE to NOT READY.

Fix: use `-P` (git is built with PCRE here — it works), or drop `\b` and post-filter with `awk`, which §4b already does for subjects. Either way, please add a one-line preflight so this can never fail silently again:

```bash
git -C "$p" log -1 -P --grep=. >/dev/null 2>&1 || echo "no PCRE support — falling back"
```

Worth noting because the skill's own CI runs `doc-lint-test` on both `macos-latest` and `ubuntu-latest`: this is exactly the kind of divergence that only bites the Mac users.

## Smaller things

1. **§6's example contradicts §5.** DOCS-6085 / MDEV-38983 is shown as **NOT READY**, but its status is `In Review`, which is in `candidate_mdev_statuses` — §5 makes that **CANDIDATE**. Since the table is the worked example an agent will pattern-match on, it should agree with the rule.
2. **`sources[server]` isn't the schema.** `.claude/doc-sources.local.json` holds `sources` as an *array* of `{product, path, ref}` (see `doc-from-ticket` §Schema). `p=<sources[server].path>` invites a map lookup that doesn't exist; say "the entry whose `product` is `server`".
3. **Four config knobs are declared but never honored.** `require_resolution`, `require_numbered_fixversion`, `hold_status` and `target_status` appear in `triage-epic.local.json`, but §5's tier table hardcodes resolution + numbered fixversion, and §2's JQL hardcodes `status = "On Hold"`. Either wire them in or drop them — a knob that does nothing is worse than no knob.
4. **Two illustrative commit counts don't reproduce.** On `preview-13.1-preview` (tip `91f37a32f3e`), MDEV-30645 matches **4 commits by message and 4 by subject**, not 3 — all four subjects begin `MDEV-30645:`. And MDEV-38975's newest matching commit there is titled `Fix CI regressions from MDEV-38975 forward-port to main`, not `MDEV-40591 …`. The *rule* is right and that "Fix CI regressions…" subject illustrates it perfectly — the numbers just need re-deriving against a current clone.
5. **`allowed-tools` includes `Write`**, while the guardrails end with "writes no `.md` file, no fact-check report, no branch, no commit". Drop `Write` and let the tool list enforce the promise.
6. Nit: `$3 ~ /^(Revert )?MDEV-30645/` also matches `MDEV-306450`. Anchor the tail — `/^(Revert )?MDEV-30645([^0-9]|$)/`.

None of 1–6 is structural. Fix the `\b` portability issue and reconcile the DOCS-6085 row, and I'd be happy to see this merged.

PR #1050, opened 2026-09-09
Remove bash call from generate_plugin_api_docs.sh

gkodinov:
Addendum to MDBF-1241: Fixed the workflow failure from executing bash in bash

### DIFF
--- a/.github/workflows/sync-server-api-plugin-docs.yml
+++ b/.github/workflows/sync-server-api-plugin-docs.yml
@@ -59,7 +59,7 @@ jobs:
             --volume "$dest:/out" \
             --workdir /src/include/mysql \
             ghcr.io/mariadb/mariadb-doc-gen:latest \
-            bash -euxo pipefail -c 'bash generate_plugin_api_docs.sh /out'
+            bash -euxo pipefail -c 'generate_plugin_api_docs.sh /out'
           test -d "$dest/md"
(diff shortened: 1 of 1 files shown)


gkodinov:
FIxes a problem with https://github.com/mariadb-corporation/mariadb-docs/pull/1012

.github/workflows/sync-server-api-plugin-docs.yml
```diff
@@ -58,8 +58,12 @@ jobs:
             --volume "$PWD/.server-src:/src:ro" \
             --volume "$dest:/out" \
             --workdir /src/include/mysql \
+            --entrypoint bash \
             ghcr.io/mariadb/mariadb-doc-gen:latest \
-            bash -euxo pipefail -c 'generate_plugin_api_docs.sh /out'
+            -euxo pipefail -c 'bash generate_plugin_api_docs.sh /out'
```
gkodinov, line 63: don't do that please.

.github/workflows/sync-server-api-plugin-docs.yml
```diff
@@ -58,8 +58,12 @@ jobs:
...
+          # The container writes as root into the bind mount, so the runner
+          # cannot delete or copy the tree afterwards without this.
+          sudo chown -R "$(id -u):$(id -g)" "$dest"
```
gkodinov, line 66: this can stay

PR #1091, opened 2026-09-16
Add Tabularis to Graphical and Enhanced Clients

debba:
Adds a page for Tabularis to the Graphical and Enhanced Clients section, with the
content-ref in the section README and the nav entry in `server/SUMMARY.md`, both
placed alphabetically after TablePlus.

Tabularis is an open source (Apache 2.0) desktop SQL workspace for Windows, macOS
and Linux. MariaDB is one of its three built-in drivers, so no plugin is involved.
I am the author.

The page follows the existing third-party pages as its template (info hint, key
features, supported databases, license footer). The MariaDB-specific part comes
straight from the driver code rather than from marketing copy: the driver detects
MariaDB from the version string and uses `EXPLAIN FORMAT=JSON` for estimated plans
and `ANALYZE FORMAT=JSON` (10.1+) when actual row counts are wanted, and tables
created `WITH SYSTEM VERSIONING` are handled in the schema explorer.

The screenshot is the Visual EXPLAIN view on an analyzed plan, which is the
feature most relevant to a MariaDB reader. I kept the Marketo form block because
the sibling pages have it; happy to drop it if that is added on your side.

This contribution is donated under the MariaDB Contributor Agreement.


mariadb-stefan-hinz:
Reviewed and merging — thank you, and thank you for the unusually careful PR description.

I spot-checked the MariaDB-specific claims against the Tabularis source rather than taking them on trust, and they hold up: the version-string detection and the `supports_analyze_format` gate in `drivers/mysql/explain.rs`, and the `table_type IN ('BASE TABLE', 'SYSTEM VERSIONED')` query in `drivers/mysql/mod.rs` — the last one with a comment explaining exactly why system-versioned tables would otherwise disappear from the list. That is the kind of detail these pages usually lack.

I pushed one small commit of house-style fixes rather than sending the PR back for them:

* Title Case on the h2 headings. This is our convention but nothing in CI enforces it, so it is easy to miss — and the two pages you used as a template (LibreDB Studio, AdminNeo) had it wrong too, which is presumably where it came from. I corrected those two in the same commit so the section is consistent. The heading slugs are unchanged, so no links move.
* `recognised` to `recognized`; the docs are en-US.
* Dropped "(MariaDB 10.1 and later)" from the `ANALYZE FORMAT=JSON` sentence. The statement is accurate and your driver does gate on 10.1, but 10.1 has been out of maintenance for years and our own `ANALYZE FORMAT=JSON` page carries no such annotation — in our docs a version note that old only dates the page. Nothing about your gate needs to change.

Keeping the Marketo form block was the right call; 39 of the 40 pages in that section have it.

The GitBook preview checks do not run on PRs from forks, so the four you may have expected to see are absent by design rather than failing. The nine that do run are all green.

Issue #996, opened 2026-08-31
Update DuckDB docs


drrtuy:
Engine has gamma maturity.
The current docs obviously were generated from the engine docs section. The docs must be re-generated using the engine's actual docs section documents.  
Installation must refer to the release 11.4,11.8,12.3 and 13.0,13.1 packages.

PR #1094, opened 2026-09-17
DOCS-6561 Rewrite the DuckDB storage engine page for gamma maturity

mariadb-tauseefkhan:
Closes DOCS-6561 / #996.

The page still described an alpha-maturity engine that had to be installed from unsigned ci.mariadb.org builds. It is regenerated from the engine's own docs section (`storage/duckdb/README.md` and `storage/duckdb/docs/`), with every claim verified against the source at tag `mariadb-11.4.13`, the engine's MTR results, and the published package repositories.

Installation now refers to the released packages, as the ticket asked: `mariadb-plugin-duckdb` (DEB) and `MariaDB-duckdb-engine` (RPM), confirmed present for 11.4.13, 11.8.9, 12.3.3, 13.0.2, and 13.1.1 on x86_64 and aarch64.

### Corrections to the old page

| Old page | Now |
|---|---|
| alpha maturity, `plugin-maturity=alpha` required | gamma; loads on a stable server with only a log warning, since the default `plugin_maturity` is one level below the server's |
| not in mainline releases, install from CI | released plugin packages, RPM and DEB |
| DuckDB 1.5.2 | 1.5.5 |
| `duckdb_memory_limit` default 1 GiB, "raise it" | default `0`, auto-sized to 80% of RAM minus `innodb_buffer_pool_size_max` |
| no `AUTO_INCREMENT`, no `UUID` | both supported; `AUTO_INCREMENT` is `CREATE TABLE` only, backed by a `mdb_autoinc_<table>` sequence |
| `DECIMAL` over 38 digits fails at DDL | stored as `DOUBLE`, or clamped when `duckdb_use_double_for_decimal=OFF` |
| non-UTF8 charsets fall back to binary comparison | rejected at DDL |
| Rocky, Fedora, openSUSE | RHEL family 8/9/10, Debian 11-13, Ubuntu 22.04/24.04/26.04 |

New sections: availability and requirements, data type and collation mapping, transactions and concurrency, `run_in_duckdb()` with its security model, database file maintenance, the full system and status variable lists, and query pushdown limitations.

### Three things I would like checked

Each of these contradicts a statement in the engine's own docs, so I may have it wrong.

1. **Concurrent writers.** `docs/shared-resources.md:56` says writes to a table from multiple connections are serialized by an internal write lock ("single-writer model in DuckDB"), and line 58 credits a MariaDB `THR_LOCK` in `Duckdb_share`. Upstream DuckDB documents the opposite within one process: concurrent writes succeed absent a conflict, appends never conflict, and optimistic concurrency control fails the second transaction to edit the same row. `ha_duckdb.cc:1107-1113` also has `lock_count()` return 0 and `store_lock()` register no lock data, so no server-side table lock is taken. The page now describes the MVCC and OCC model.

2. **Unenforced `PRIMARY KEY`.** `r/create_table_constraint.result` shows a table declared with a primary key and three unique indexes reaching DuckDB with one `NOT NULL` constraint and no indexes, and `t/ha_duckdb.test:31` inserts a duplicate primary key row on purpose. The page states plainly that a `PRIMARY KEY` or `UNIQUE` index is not enforced and that a row-level `UPDATE` or `DELETE` then affects both rows, since `dml_convertor.cc:217-259` builds its `WHERE` from the first key's columns. That seems worth documenting, but confirm it is intended rather than a bug.

3. **"DuckDB has no indexes"** (`docs/architecture.md`) — DuckDB does have ART indexes. The page says instead that the engine creates none and reports no index capabilities to the optimizer.

Page also notes that invisible columns are unsupported, which is the released behavior; MDEV-40846 is on branch `11.4` but not in any release tag yet.


mariadb-tauseefkhan:
@drrtuy this is the rewrite for #996 — would you mind casting an eye over the technical side? Github wont let me add you as a reviewer on this repo, hence the ping.

Rendered preview, easier than reading the diff:
https://mariadb.com/docs/server/~/revisions/PweVATyqJGHpv12MVxPz/server-usage/storage-engines/duckdb-storage-engine

The three questions at the end of the description are the ones I'd most like your take on — each is a place where I ended up saying the opposite of `storage/duckdb/docs`, so I could easily be the one whos wrong. The unenforced PRIMARY KEY in particular: `t/ha_duckdb.test:31` looks deliberate, but I wanted to be sure before documenting it as expected behaviour.

cc @stefanhinz (you create the doc)

PR #1097, opened 2026-09-17
DOCS-6654 Note that Enterprise Server 12.3 does not install stunnel

dbart:
Closes [DOCS-6654](https://mariadbcorp.atlassian.net/browse/DOCS-6654) (reported by Ralf Gebhardt).

Galera rsync SST encrypts transfers with `stunnel`, but **Enterprise Server 12.3 packages don't install it** — neither as a requirement nor as a recommendation, on RPM or DEB. The fix was deliberately deferred past the 12.3 ES beta, so the workaround belongs in the docs.

Two warnings added:

| Page | Placement |
|---|---|
| `introduction-to-state-snapshot-transfers-ssts.md` | In the rsync/stunnel tab that already says *"Be sure to have `stunnel` installed"* — where a reader configuring encrypted rsync SST lands |
| `mariadb-enterprise-cluster-security.md` | After the SST TLS modes table — none of those modes can be satisfied on 12.3 ES until the binary exists |

### ⚠️ One deliberate departure from the ticket

**The ticket asks for the same note on Community Server 12.3. Source doesn't support that**, so I narrowed it rather than repeating the ES wording.

CS 12.3 lists `stunnel` as a `Recommends` of `mariadb-server-galera` on both packaging paths, so a default install *does* get it — apt installs Recommends, dnf installs weak deps. It's absent only under `--no-install-recommends` or `install_weak_deps=False`. The CS paragraph documents that narrower condition. **Ralf, this is the bit worth your eye.**

### Verification

| Claim | Evidence |
|---|---|
| ES 12.3 ships no `stunnel` at all | zero matches across `debian/`, `cmake/` @ `7927a5a1` (`origin/12.3-enterprise`) |
| Regression from ES 11.8 | `cmake/cpack_rpm.cmake:276`, `debian/control:721` @ `1eae6f8f` |
| CS 12.3 has it as `Recommends`, not `Depends` | `cmake/cpack_rpm.cmake:276`, `debian/control:985-987` @ `0fcc374c` (`origin/12.3`) |
| `mariabackup` unaffected | `stunnel` appears only in `scripts/wsrep_sst_rsync.sh`; 0 matches in `scripts/wsrep_sst_mariabackup.sh` @ `7927a5a1` |

Full 8-claim fact-check report is posted to the Jira ticket.

### Two things deliberately left out

- **No link to the upstream ticket.** It's an internal Enterprise project: anonymous API access returns **401**, while its `/browse/` URL returns 200 because Jira Server serves a login page — so lychee would pass it in CI while readers hit a wall.
- **No "fixed in" wording.** The upstream bug is still Open, so anything forward-looking would be unverifiable.

### Follow-up filed separately

[DOCS-6671](https://mariadbcorp.atlassian.net/browse/DOCS-6671) — the MDEV-40630 release-note text published across five Community Server series calls `stunnel` "a required package of MariaDB-server-galera", which the same source contradicts. Out of scope here.

`.claude/hooks/doc-lint.sh` passes on both changed files.

[DOCS-6654]: https://mariadbcorp.atlassian.net/browse/DOCS-6654?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[DOCS-6671]: https://mariadbcorp.atlassian.net/browse/DOCS-6671?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


dbart (commented):
From Ralf: I am not sure myself, we maybe should be more specific to say “MariaDB Enterprise Server 12.3.3-0 (beta)” as we want to get this fixed for the GA.

PR #1100, opened 2026-09-17
DOCS-6586: gate orphaned and gutted pages in CI, via a checked-in allowlist

mariadb-stefan-hinz:
Completes **DOCS-6586**. Parts 1 and 3 landed on 2026-09-04 (`3e0752d16`, `71f3ac501`); the two remaining gates were blocked on the acknowledgment-path decision, which Daniel made on the ticket on 2026-09-14 — a checked-in allowlist file, with the orphan half self-pruning. Both are implemented as specified, including the self-check.

**Every check in `doc-lint.sh` except one now has a CI counterpart.** The exception is the retired-Knowledge-Base link rule, which is still local-only — corrected in the contributor doc in the review-fix commit below, where the original sentence was wrong.

## The register — `.claude/hooks/doc-lint-allow.yml`

```yaml
orphan:
  - path: release-notes/enterprise-server/12.3/whats-new.md
    reason: "unreleased, hidden until beta ships — DOCS-6391"

shrink:
  - path: tools/debian/README.md
    reason: "3 of 5 child pages retired — DOCS-5976"
```

- **`allowlist.py` is the only parser**, standard-library only and a strict subset of YAML: it rejects anything outside the shape above with a **line number**. PyYAML would need a pip install these checks cannot assume (they run from a contributor's pre-commit hook and from a runner that installs nothing), and a `try: import yaml` fallback would be two parsers that can disagree about one file. The accepted shape is real YAML — cross-checked against Ruby's parser, which reads the same structure.
- **It refuses `all`.** The `DOC_LINT_ALLOW_*` variables still take it for a one-off local run; in a checked-in file it is a permanent repo-wide disable that reads like an acknowledgment.
- **A stale entry fails, naming the line to delete.** An `orphan:` entry whose page has since been listed in `SUMMARY.md` (your condition), or whose page is gone, or which never named a page in any space; a `shrink:` entry whose page is gone. Neither a narrow file scope nor a local `DOC_LINT_ALLOW_*=all` can hide one, because a stale entry is a defect in the register rather than a claim about a page. There is no "un-shrink" signal, so a `shrink:` entry whose page has since regrown still needs the occasional manual pass — as you predicted.
- **A malformed register is exit 2, never an empty one.** Read as empty it would fail a PR that *had* acknowledged its finding correctly, leaving the author a valid-looking entry in the diff and a red gate with no explanation.
- The register ships **empty**, so both gates start green.

## The two gates

| Workflow | Scope | Why |
|---|---|---|
| `navcheck-pr.yml` | tree-wide, history-aware | Deleting a nav entry orphans a page the diff need not contain — the diff holds `SUMMARY.md`, not the page. And the stale audit has to run on a PR that touches no Markdown at all, since an entry goes stale when some *other* PR lists or deletes the page it names. The whole run is milliseconds. |
| `shrinkcheck-pr.yml` | the changed set | The distinction that decides scope in this repo is whether a finding is **local to the file that changed**. A dead include is not — one shared target breaks 20 pages — so that gate is tree-wide. Lost lines are. |

Both **assert the base commit is present before invoking**, because both scripts treat an unresolvable base as a SKIP and exit 0 — right for a pre-commit hook, silently green for a gate. Both **assert non-vacuity**: navcheck's clean line now carries the page count and the workflow requires it nonzero; shrinkcheck reports what it received and the workflow requires that to equal the number of paths it was fed.

## The extraction

The shrink guard moved out of `doc-lint.sh` into `shrinkcheck.py` — the move `includecheck.sh` made in part 1, and for the same reason: the workflow must not route through `doc-lint.sh`, which would re-run codespell, lychee and the orphan guard. It is **Python rather than the bash it came from** so the register has one parser and not two; that makes the local guard depend on python3 where it needed only bash, acceptable only because it now *has* a CI counterpart, which is the trade `fragcheck.py` and `navcheck.py` already make.

## Proving the firing path

- **Suite 61 → 93 cases**, 0 skipped, and **every new case proven red by mutation** — 11 mutations, all caught: allowing `all` in the file, dropping the reason requirement, skipping the stale audit, reading the file but not the env (and vice versa), reading a malformed register as empty, dropping either counts line, splitting stdin on whitespace instead of NUL, letting `ALLOW_SHRINK=all` hide a stale entry, and turning a missing base into a pass.
- **Replayed against `c6ea5549a`** — the DOCS-6442 commit this guard exists for — `shrinkcheck.py` names exactly one file out of 45 compared, at 93% (276 of 298 lines), matching the original measurement.
- **Every rendered `run:` block extracted with `ruby -ryaml` and executed locally**, on the green path, on findings, on a malformed register and on a stale entry.

Two defects found that way, both by running the thing rather than reading it:

1. **`navcheck-pr.yml`'s non-vacuity `sed` matched nothing.** It anchored the count to an opening parenthesis, which is where `check` puts it and `new` does not (there it follows `unchanged; `). The count came back empty and the step would have **failed every green run**. Same method that found part 1's `xargs` defect and DOCS-6587's SIGPIPE one.
2. **Three strings had to be reworded** for codespell (`accreting` → `accrediting`) and shellcheck **SC2016** (backticks inside single quotes, in a case title and an assertion) — the CI job runs at default severity, so both would have turned it red. The SC2016 class is the third time it has come up here; #1028 hit it in three case titles.

## Test plan

```
doc-lint-test.sh                              93 passed, 0 failed, 0 skipped
shellcheck (all tracked .sh)                  clean
actionlint (all 16 workflows)                 clean
codespell (CI-exact, changed files)           clean
includecheck --stdin0 (tree-wide)             4,295 includes across 9,744 files, all resolve
fragcheck new HEAD                            no new dead anchors, no new stolen ids
navcheck new HEAD                              no newly orphaned pages (44 pre-existing)
allowlist.py validate                         0 orphan entries, 0 shrink entries
```

## Two things for the record

- **The pre-existing orphan backlog is 44 today**, where the ticket says 219 and its 2026-09-04 comment says 209. Pages have been nav-listed since; not a regression. The cookbook and the skill now both say to take the figure fresh rather than quote it.
- `fragcheck check` counts an `href="#x"` **inside an inline code span** as a fragment link, so the DOCS-6492 example in `cookbook-pre-pr.md` reports as one dead anchor. Pre-existing, one of the 16, and a false-positive class rather than a defect in that page — left alone rather than reworded to dodge a linter.

## Review fixes (`44faab5c8`, after approval)

Daniel's review found one false claim and five latent gaps; all six are in, rather than parked.

- **Finding 1, the blocker.** `cookbook-pre-pr.md` told contributors that a local finding is always a finding CI repeats, while `doc-lint.sh:367` says the retired-Knowledge-Base rule has **no** CI counterpart. The doc now names that one check and says it runs only locally; the suite header carried the same sentence and is corrected too. Writing the rule out tripped the rule — the page now spells the host and the path separately, which is recorded in the sentence itself.
- **Finding 2.** `shrinkcheck.py` tested paths for eligibility against the CWD and read them against the computed repo root, so a run from a subdirectory measured a *different* file of the same relative name and printed a confident count for it. Paths are re-anchored to the root; one outside the repo is exit 2. This was the one gap the workflow's own assertions could not catch — it asserts `given` and `eligible`, never `compared`.
- **Finding 3.** The stale-entry audit now runs above the two `rev-parse` guards, where `navcheck.py` already had it.
- **Finding 4.** `git()` answers `not ok` instead of raising `FileNotFoundError` when git is off `PATH`, and the `shrinkcheck` block in `doc-lint.sh` regained the `command -v git` guard its navcheck sibling has.
- **Finding 5.** A `reason` must now be indented deeper than the `- ` opening its entry: at column 0, YAML makes it a separate top-level key, so the entry would carry no reason at all.
- **Finding 6.** 94% → 93% in the `shrinkcheck.py` header; the alphabetical-order line in the register says it is a convention the parser does not enforce; the 162-character line is rewrapped.

Not taken: the `setup-python` v5/v7 split is pre-existing and includes `fragcheck-pr.yml`, so it is not this PR's to change.

**Suite 88 → 93**, one case per fix, each proven red first: the subdirectory measurement, an outside path, a stale entry under an unresolvable base, git off `PATH`, and the column-0 reason. shellcheck and actionlint clean, register still 0 entries, backlog still 44.

## The gate's first real catch, and a defect it exposed (`f71cbdd22`)

The gutted-page gate failed on this very PR, and the finding was not this PR's: `release-notes/maxscale/25.10/README.md`, 57 lines → 9, gutted on `main` by `e8a2b23d0`. All five MaxScale version landing pages lost their content-ref lists the same way, so that is an upstream pattern change rather than an accident — reported, not fixed here.

**Why it landed on us is a defect in all three history-aware gates.** `actions/checkout` on a `pull_request` checks out `refs/pull/N/merge`, so HEAD is a merge of the PR head into the base branch *as it stands now*, while `github.event.pull_request.base.sha` is a snapshot of that branch from when the PR was opened or last synchronized. Pairing the two spans every commit the base branch gained in between — here, 23 paths in scope instead of 13, and a page somebody else shrank reported against an author who cannot change it.

The merge commit's **first parent** is the base tip that was merged, so `HEAD^1..HEAD` is exactly what the PR does to the base branch. Each gate now resolves that in its own step and every consumer reads the resolved value, with `base.sha` as the fallback for a head-ref checkout or a local re-run; the presence assertion is unchanged. `fragcheck-pr.yml` gets the same fix although it shipped in DOCS-6524 — the defect is identical there and fails in the same direction, an anchor broken on `main` after the branch point reported as the PR's.

Verified on this PR's own merge commit `cd5a518bd`: resolved base `b8dca955b`, **3** Markdown paths in scope instead of 23, shrinkcheck 0 gutted, navcheck clean at 9,498 pages.


dbart (commented):
## Review

The implementation matches the acknowledgment-path decision on DOCS-6586 verbatim — one file, two sections, `path` + `reason`, the orphan half self-pruning — and the evidence in the PR body holds up under re-running. One checked-in claim is false and worth fixing before merge; findings 2–6 are latent robustness gaps rather than blockers, and none of them stops either gate doing its job.

### Re-verified independently, not just read

| Claim | Result |
|---|---|
| Suite 88 cases, 0 skipped | ✅ `88 passed, 0 failed, 0 skipped` |
| shellcheck + actionlint clean | ✅ both exit 0 (the 7 findings are in `release-notes/maxscale/script/*.sh`, excluded in CI) |
| `c6ea5549a` replay: 1 of 45 at 93% | ✅ exactly — `server/server-usage/storage-engines/README.md`, 276 of 298 lines |
| navcheck non-vacuity `sed` | ✅ extracts `9491` from the real clean line; the `[^0-9]` anchor is load-bearing and correct |
| shrinkcheck `given` assertion | ✅ extracts `46` from the replay log; the `git diff -z` NUL count equals `received` |
| Register starts green; 44 pre-existing orphans | ✅ `0 orphan entries, 0 shrink entries`; `44 pre-existing, 9491 pages in 12 spaces` |
| The new `§ 1a` anchor | ✅ fragcheck resolves it; the 1 dead anchor is the pre-existing `#x` code-span false positive at line 112 |

### 1. `dev-docs/cookbook-pre-pr.md:23` — the headline claim is false (worth fixing before merge)

> **every check `doc-lint.sh` runs now has a CI counterpart**, so a finding on your machine is a finding CI will repeat, and none of them is local-only any more.

The retired-Knowledge-Base-link check is still local-only: `doc-lint.sh:361` sets `rc=1` on a `mariadb.com/kb/` hit, has its own `DOC_LINT_ALLOW_KB` hatch, and no workflow references it (grepped `.github/workflows/` — nothing). The comment on that same line says "NO CI counterpart (**the last one**)", so the code and the contributor doc contradict each other inside one PR.

This is the one I'd fix before merge, because it tells a contributor that a local KB finding will be repeated by CI. "every check *that has a CI counterpart*…", or naming the four delegated checks, would be accurate. The same sentence leads the PR body and the 2026-09-17 ticket comment, so it is worth correcting in all three.

### 2. `shrinkcheck.py:222` vs `:151` — paths are eligibility-tested against the CWD but read against the computed repo root

When the two differ the guard does not skip or error: it measures a *different file* and reports success. Repro with no commits, working-tree edit only, `server/README.md` gutted 111 → 1 line:

```
from repo root:  --base HEAD server/README.md → "possible gutted page, lost 110 of 111 (99%)"  rc=1
from server/:    --base HEAD README.md        → "1 file(s) compared, 0 possibly gutted"        rc=0
```

The second run silently measured the repo-root `README.md` (7 lines, under the floor) and printed a confident count for it.

Not reachable today — CI runs from the root, `pre-commit.sh:46` does `cd "${CLAUDE_PROJECT_DIR:-.}"`, and every documented invocation is root-relative — so this is a latent trap rather than a live hole. But it is the one gap in this PR's own "a check that cannot run must not report success" thesis: the workflow asserts `given` and `eligible` and never `compared`, so the assertion would not catch it. `navcheck.py` already guards this class with its `show-toplevel` comparison from #1028; `shrinkcheck.py` has no equivalent.

### 3. `shrinkcheck.py:249–261` — the stale audit is skipped on an unresolvable base, where `navcheck.py` runs it anyway

Same stale entry in both:

```
shrinkcheck --base no-such-rev → "base revision not found — SKIPPED"   rc=0   (stale entry invisible)
navcheck new no-such-rev       → "1 stale acknowledgment … line 37"    rc=1
```

A stale entry is a file-existence test that needs no git at all, and the PR body's own reasoning — "a stale entry is a defect in the register rather than a claim about a page" — argues it should survive a missing base. Moving the audit above the two `rev-parse` guards makes the two scripts agree. Not live: the workflow asserts the base first.

### 4. `doc-lint.sh:352` — the shrinkcheck block dropped the `command -v git` guard its navcheck sibling still has at `:263`

With git off `PATH`, `shrinkcheck.py:101` raises an uncaught `FileNotFoundError` traceback and exits 1, where the bash it replaced skipped. The suite's "tree with no git" case (`doc-lint-test.sh:1303`) `git archive`s into a bare temp dir, so it covers a non-work-tree rather than a missing binary and this path is untested. Near-unreachable inside a git pre-commit hook, but a `try/except` in `git()` would restore the SKIP contract the header documents.

### 5. `allowlist.py:210–221` — a column-0 `reason:` is attached to the preceding entry, where YAML makes it a separate top-level key

```yaml
orphan:
  - path: server/foo.md
reason: "not indented under the entry"
```

`allowlist.py` reads this as `1 orphan entry` carrying that reason. Ruby's YAML reads `{"orphan"=>[{"path"=>"server/foo.md"}], "reason"=>"..."}` — the entry has **no** reason. So the "every entry needs a reason" rule, which is the point of the register, can be satisfied by something that is not structurally a reason, and the header's "the file is valid YAML, so editors and GitHub still highlight it; it is just a subset" stops holding for that shape. Requiring the `reason` line to be indented deeper than its `- ` closes it.

### 6. Minor

- `shrinkcheck.py:26` says the `c6ea5549a` list comes out "at 94%". It is 93%, as both the workflow header and the PR body say — a figure carried over from the bash original.
- `doc-lint-allow.yml:34` asks for alphabetical order that nothing enforces, though the parser already rejects duplicates.
- `cookbook-pre-pr.md:142` is a 162-character line where the file otherwise wraps at ~100.
- `navcheck-pr.yml:81` and `shrinkcheck-pr.yml:79` pin `setup-python@v5` while four other workflows use `@v7`. Pre-existing split (fragcheck is v5 too), not introduced here.

Both gates start green, both fire on the cases they exist for, and the firing paths are genuinely proven rather than asserted — the `navcheck-pr.yml` `sed` defect you caught by executing the rendered block is exactly the class that reading would have missed.

dbart (approved):

PR #1103, opened 2026-09-18
Referred to BOOLEAN

Friedrich2:


mariadb-stefan-hinz:
Merged — thanks @Friedrich2, the note is a good addition and the claim holds up.

**Fact-checked rather than taken on trust:** `BOOL_SYM` and `BOOLEAN_SYM` both resolve to `set_handler_length(&type_handler_stiny, 1)` in the server's `sql/sql_yacc.yy`, i.e. `TINYINT` with display width 1, and a live 13.0.2 server reports `flag tinyint(1)` for a `BOOLEAN` column.

I amended the PR (edits by maintainers was on) rather than sending it back for wording:

* "as more speaking alias for TINYINT(1)" → the sentence now says the `BOOLEAN` column "names the intent more clearly"; `TINYINT(1)` is code-spanned, and "data type" is two words in house style.
* The link pointed at the published `mariadb.com/docs/...` URL. `BOOLEAN` sits in the same GitBook space, so it is a relative path — `../numeric-data-types/boolean.md`, the form `tinyint.md` already uses for the reverse link.
* Dropped the doubled blank line.
* Added what the substitution costs, because the bullet your note qualifies sells `CHAR(0)` on storage: `TINYINT(1)` holds a one-byte value where a `CHAR(0) NULL` column carries its state in the row's `NULL` bit alone.

**Your note also turned up a false claim two paragraphs up, now fixed in the same PR.** The page said `VARCHAR(0)` columns "cannot be part of an index". In `sql/sql_table.cc` a zero-length key part is rejected **only when the column is `NOT NULL`** — so a nullable `VARCHAR(0)` or `CHAR(0)` indexes fine. Verified on 13.0.2 (both `UNIQUE KEY` and a plain `KEY` are accepted; the `NOT NULL` form fails with `ERROR 1167 (42000): The storage engine InnoDB can't index column`) and the same condition is in the 10.6 tree, so it holds across every supported series. That matters for the "Row Marking" bullet right below it: the one-active-row pattern needs a unique index on a nullable column, which the old sentence said was impossible.

12/12 checks green, rebase-merged as `b854f7b87` + `d397411fb`. Your `patch-2` branch is yours to delete.

Friedrich2:
@mariadb-stefan-hinz  Many thanks for your insightful and impressive changes!

PR #1115, opened 2026-09-22
Connect: not deprecated

grooverdan:


mariadb-stefan-hinz:
Thanks, Daniel! I looked at this PR only after handling an Issue complaining about the exact same goof-up. So, I already fixed it (see https://github.com/mariadb-corporation/mariadb-docs/pull/1123), and will close this PR without merging. Thanks again!

